PR #10, opened 2019-03-21
Add support for puppeteer debugging options

abrom:
Also update the README to expand on direct use of Grover and caveats regarding relative paths


abrom:
FYI @Richacinas

PR #13, opened 2019-05-01
screenshot support

ss44:
This is a rudimentary request, I figure there's a bunch of code normalization that can occur first but would rather have some feedback on how you'd really want to approach this feature. 

While there are a few gems that handle screenshots they currently rely on phantomjs or webkit both of which are growing long in the tooth when it comes to modern css and JS support.

this is kind of my attempt to get the ball rolling on inclusion of this logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,9 @@ grover = Grover.new('https://google.com', format: 'A4')
 # Get an inline PDF
 pdf = grover.to_pdf
 
+# get a screenshot
+png = grover.screenshot
+
 # Options can be provided through meta tags
 Grover.new('<html><head><meta name="grover-page_ranges" content="1-3"')
 Grover.new('<html><head><meta name="grover-margin-top" content="10px"')
--- a/lib/grover.rb
+++ b/lib/grover.rb
@@ -89,7 +89,75 @@ class Grover
         }
       }
     FUNCTION
+
+    method :screenshot, Utils.strip_heredoc(<<-FUNCTION)
+      async (url_or_html, options) => {
+        let browser;
+        try {
+          let launchParams = #{launch_params};
+
+          // Configure puppeteer debugging options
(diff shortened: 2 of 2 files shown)


abrom:
Nice one @ss44 👍

Yeah I think there would be some likely refactoring needed to remove some of the duplication. Would likely also need to support optionally calling to `page.viewport` (https://github.com/GoogleChrome/puppeteer/blob/v1.15.0/docs/api.md#pageviewport)

There might also be the potential to modify the middleware to support handling of say `.png` and `.jpg`/`.jpeg` extensions too. ie allow rack requests to just add those extensions to have an image returned instead of the raw HTML. Hmm :) Although happy to see that as a separate change!

I'd like to see some spec tests to go along with this sort of change. It might be that a test pattern is rendered in HTML then we could just test the resulting image (PNG/JPEG) for the expected colours as the expected pixel offsets. 

One concern would also be how this is configured. Given that people may want to configure the image saving differently to the PDFs (when configured via `Grover.configure`). Might require allowing 'global' configurations as well as conversion type specific options. Hmm. Also happy to see that as a separate change

ss44:
@abrom awesome thanks for the feedback. 

I'll try and tackle cleaning up the duplicated blocks and the rspec tests, later this week and get the checks passing. 

Thanks for putting this gem together. Found it super helpful in getting going and codes super clean.

abrom:
Hey @ss44, any progress on this?

ss44:
sorry I've been sidetracked by work and just life lately. I will revisit this soon as I can just haven't had time to fix this up. sorry for the delay.

abrom:
FYI @ss44 I've implemented this in https://github.com/Studiosity/grover/pull/25

Doesn't include middleware just yet but i've included the `screenshot` method as well as `to_png` and `to_jpeg` helper methods.

I'll close this PR as the function is now in master.

PR #19, opened 2019-07-01
Add sample code to render a Rails view as PDF

soundasleep:
Not sure if this is the right place to include these docs, or even if this should maybe be a method that's inserted into all Controllers (maybe through the middleware)?

I was experimenting with `wicked_pdf`, and it's pretty easy with that project to render a PDF of _any_ Rails view. This code does the same thing with Grover, it could be useful for others?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,19 @@ N.B.
 * options are underscore case, and sub-options separated with a dash
 * all options can be overwritten, including `emulate_media` and `display_url`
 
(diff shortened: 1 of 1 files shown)


abrom:
Regarding the spec test failures, leave those with me.. looks like something with the `pdf_reader` gem may have changed.. hmm :/

soundasleep:
Done! Thank you! :)

PR #29, opened 2019-09-18
Add configuration option `args`

joergschiller:
Useful for adding custom Chromium command line switches (e.g. `--font-render-hinting=none`).

### DIFF
--- a/lib/grover.rb
+++ b/lib/grover.rb
@@ -23,7 +23,7 @@ class Grover
     dependencies puppeteer: 'puppeteer'
 
     def self.launch_params
-      ENV['GROVER_NO_SANDBOX'] == 'true' ? "{args: ['--no-sandbox', '--disable-setuid-sandbox']}" : '{}'
+      ENV['GROVER_NO_SANDBOX'] == 'true' ? "{args: ['--no-sandbox', '--disable-setuid-sandbox']}" : '{args: []}'
     end
 
     def self.convert_function(convert_action)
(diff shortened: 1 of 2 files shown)


joergschiller:
Hi! First I want to thank you for this great project. We're in the progress switching from [Wicked PDF](https://github.com/mileszs/wicked_pdf) (using wkhtmltopdf) to Grover.

Grover makes this pretty easy. We only had some issues with font rendering on Linux which can be fixed with an additional command line argument to chromium (`--font-render-hinting=none`).

I had problems thinking about a good rspec for this case. The final PDF doesn't really change in terms of content. It's just a problem with subpixel positioning.

lib/grover.rb
```diff
@@ -23,7 +23,7 @@ class Processor < Schmooze::Base
     dependencies puppeteer: 'puppeteer'
 
     def self.launch_params
-      ENV['GROVER_NO_SANDBOX'] == 'true' ? "{args: ['--no-sandbox', '--disable-setuid-sandbox']}" : '{}'
+      ENV['GROVER_NO_SANDBOX'] == 'true' ? "{args: ['--no-sandbox', '--disable-setuid-sandbox']}" : '{args: []}'
```
abrom: not sure this change is necessary if you instead just assign the args on L46 ?? Why have a default and concat if they're always empty?

abrom:
Hi @joergschiller, any thoughts on my comments?

abrom:
Hi @joergschiller didn't get any replies from you so went ahead and fixed the issues raised in another PR: #31 (released in v0.8.2)

Going to close this one

Issue #40, opened 2020-01-15
cover_path params not present in Rails Controller


jnimety:
I might be doing something wrong here, but I am unable to see my cover_path params when the request is initiated by grover. When I request the page directly via browser the params are visible.
Is this a known limitation?

`front_cover_path: pdf_cover_page_path(title: "Title", subtitle: "Subtitle"),`

```
[5, 16] in app/controllers/pdf_cover_pages_controller.rb
    5:   def show
    6:     byebug
=>  7:     @title = pdf_cover_page_params[:title]
    8:     @subtitle = pdf_cover_page_params[:subtitle]
    9:   end
   10:
   11:   private
   12:
   13:   def pdf_cover_page_params
   14:     params.permit(:title, :subtitle)
   15:   end
   16: end
(byebug) request.params
{"controller"=>"pdf_cover_pages", "action"=>"show"}
(byebug) request.url
"https://host.redacted.net/pdf_cover_page?subtitle=Subtitle&title=Title"
(byebug) params
<ActionController::Parameters {"controller"=>"pdf_cover_pages", "action"=>"show"} permitted: false>
```

abrom:
No, looks like a bug to me! Rails is doing some caching of the request parameters in the `env` which results in the params being what ever they were for the original request..

However the cache can be busted with the following:

```
request.delete_header 'action_dispatch.request.parameters'
request.delete_header 'action_dispatch.request.query_parameters'
puts params
=> {"title"=>"Title", "subtitle"=>"Subtitle", "controller"=>"pdf_cover_pages", "action"=>"show"}
```

As a short term measure you could put those `delete_header` calls into your cover page controller in a `before_action` callback. 

Taking a look at other places that Rails messes about with the above headers I can see the following method in their `TestCase` class:
```
        def scrub_env!(env)
          env.delete_if { |k, v| k =~ /^(action_dispatch|rack)\.request/ }
          env.delete_if { |k, v| k =~ /^action_dispatch\.rescue/ }
          env.delete "action_dispatch.request.query_parameters"
          env.delete "action_dispatch.request.request_parameters"
          env["rack.input"] = StringIO.new
          env.delete "CONTENT_LENGTH"
          env.delete "RAW_POST_DATA"
          env
        end
```

The likely solution will be to modify Grover to take Rails' lead on what they expect to be scrubbed between requests. Leave it with me, I'll tag you when I've got a PR together

abrom:
[...]

PR #41, opened 2020-01-16
Scrub the middleware cover page request env before request

abrom:
Fixes issue raised in #40 

The original request env was adding cached parameters resulting in the cover page requests using the cache instead of generating the params themselves. This change scrubs the env before calling to the app when generating the cover page. The method used was pulled out of the ActionController TestCase class so should give us reasonable expectations over what needs to be cleaned up between requests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ Metrics/BlockLength:
     - "**/*_spec.rb"
 
 Metrics/ClassLength:
-  Max: 125
+  Max: 135
 
 Metrics/LineLength:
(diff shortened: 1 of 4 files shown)


abrom:
FYI @jnimety 

I've tested it locally and the fix works fine for me. Would you mind verifying that this branch resolves the issue for your configuration too

If it works as expected I'll release a build

jnimety:
I'll test tomorrow morning, thanks again for the quick turnaround.

jnimety:
Works fine for me as well.

PR #43, opened 2020-02-15
Ensure viewport width and height are integers

### LINKED ISSUE #42: Is it possible to set the viewport size with `<meta>` tags?
Is it possible to set the viewport size with `<meta>` tags? Ex:

```
<meta name="grover-viewport-width" content=640>
<meta name="grover-viewport-height" content=480>
```

This doesn't work because 640 and 480 are converted to strings and puppeteer throws an error "Protocol error (Emulation.setDeviceMetricsOverride): Invalid parameters width: integer value expected; height: integer value expected"

(Asking here because it comes up when searching, so hopefully that will help others benefit from the answer)

_Originally posted by @willkoehler in https://github.com/Studiosity/grover/issues/26#issuecomment-586035574_

willkoehler:
This should fix #42

### DIFF
--- a/lib/grover.rb
+++ b/lib/grover.rb
@@ -264,7 +264,8 @@ class Grover
     Utils.deep_merge! combined, meta_options unless url_source?
 
     fix_boolean_options! combined
-    fix_numeric_options! combined
+    fix_integer_options! combined
+    fix_float_options! combined
     fix_array_options! combined
 
     combined
@@ -304,11 +305,18 @@ class Grover
 
   FALSE_VALUES = [nil, false, 0, '0', 'f', 'F', 'false', 'FALSE', 'off', 'OFF'].freeze
 
-  def fix_numeric_options!(options)
-    options['viewport']['width'] = options['viewport']['width'].to_i if options.dig('viewport', 'width')
-    options['viewport']['height'] = options['viewport']['height'].to_i if options.dig('viewport', 'height')
-    options['viewport']['device_scale_factor'] = options['viewport']['device_scale_factor'].to_f if options.dig('viewport', 'device_scale_factor')
-    options['scale'] = options['scale'].to_f if options.key? 'scale'
+  def fix_integer_options!(options)
+    ['viewport.width', 'viewport.height'].each do |opt|
+      keys = opt.split('.')
+      Utils.deep_assign(options, keys, options.dig(*keys).to_i) if options.dig(*keys)
+    end
+  end
+
+  def fix_float_options!(options)
+    ['viewport.device_scale_factor', 'scale'].each do |opt|
+      keys = opt.split('.')
+      Utils.deep_assign(options, keys, options.dig(*keys).to_f) if options.dig(*keys)
+    end
   end
 
   def fix_array_options!(options)
--- a/lib/grover.rb
+++ b/lib/grover.rb
@@ -305,9 +305,10 @@ class Grover
   FALSE_VALUES = [nil, false, 0, '0', 'f', 'F', 'false', 'FALSE', 'off', 'OFF'].freeze
 
   def fix_numeric_options!(options)
-    return unless options.key? 'scale'
-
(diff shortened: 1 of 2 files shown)


willkoehler:
Looks like some Rubocop and spec problems ☹️ I'll take a look in the AM and get them resolved.

willkoehler:
Is `options['scale']` a puppeteer parameter? https://github.com/Studiosity/grover/blob/7dc5056b5de25c1a885684dcc7bfa260530614a5/lib/grover.rb#L310

I thought it was somehow mapping to the viewport `device_scale_factor`. But it looks like it's something different? In any case I added a line to convert `options['viewport']['device_scale_factor']` to numeric as well.

willkoehler:
@abrom this PR is ready, but I'm kind of at a dead-end between Rubocop and Code Climate. I refactored to make Rubocop happy and now Code Climate is failing with "Class Grover has 22 methods (exceeds 20 allowed). Consider refactoring." I don't feel like I have a good strategy to extract logic from `Grover`. Can you take a look?

Also looks like the Travis build failure is not related to my changes? See https://travis-ci.org/Studiosity/grover/jobs/650898717#L338-L340 Maybe it will pass if you trigger a rebuild?

abrom:
> Is `options['scale']` a puppeteer parameter?
> 
> https://github.com/Studiosity/grover/blob/7dc5056b5de25c1a885684dcc7bfa260530614a5/lib/grover.rb#L310
> 
> I thought it was somehow mapping to the viewport `device_scale_factor`. But it looks like it's something different? In any case I added a line to convert `options['viewport']['device_scale_factor']` to numeric as well.

Yes, it's a `pdf` option: https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagepdfoptions

abrom:
> @abrom this PR is ready, but I'm kind trapped in a corner between Rubocop and Code Climate. I refactored to make Rubocop happy and now Code Climate is failing with "Class Grover has 21 methods (exceeds 20 allowed)." I don't feel like I have a good strategy to extract logic from `Grover`. Can you take a look?
> 
> Also looks like the Travis build failure is not related to my changes? See https://travis-ci.org/Studiosity/grover/jobs/650898717#L338-L340 It seems like a flickering test. In different pushes, sometimes it passes, sometimes it fails.

The `Grover` class will need some refactoring, but I'm happy for that to happen at a later date. I suspect the first pass will be to extract out the option 'fix' code..

Yeah I've seen those before. Chromium crashes on boot occasionally. I've seen it in Travis and on a regular Linux box. There are threads floating about reporting the issue but clearly still an issue. I've kicked the builds and it's passed now.

willkoehler:
> The `Grover` class will need some refactoring, but I'm happy for that to happen at a later date. I suspect the first pass will be to extract out the option 'fix' code..

I think we could extract the functionality of `combine_options` into its own class? It seems fairly self-contained. If not, extracting the 'fix' code seems straightforward. I'm happy to take a stab at it  in a PR if you're interested in tackling it now.

abrom:
Yes, extracting out `combine_options` and the `fix..` functions seems reasonable. It would require passing in both the options and the `@url` variable as it's used the `meta_options` (`meta_tags`) and the `url_source?` helper methods (which would also get moved into the other class). So looking at basically all of the code below the `combine_options` method (except `normalized_options`).

The `normalized_options` method could get moved too, but it would need to be a part of the public interface for the new class. As would extracting out things like the `root_path`, `front_cover_path` & `back_cover_path`.

willkoehler:
Sounds good 👍  I'll take a pass at it and open a PR to get your feedback.

willkoehler:
Code Climate is at it again 😬 https://codeclimate.com/github/Studiosity/grover/pull/46 But I was actually thinking of making a helper for all the `fix_xx_options!` methods that took an array of options and a transformation method/block. Maybe I'll look into that and see if that makes Code Climate happy.

PR #50, opened 2020-04-24
Add specs to test new cookies option

abrom:


spec/grover_spec.rb
```diff
@@ -329,6 +329,14 @@
           it { expect(pdf_text_content).to eq 'Your browser made it!' }
         end
       end
+
+      context 'when passing through cookies option' do
```
abrom, line 333: FYI @Richacinas here is a simple test that validates that your new `cookies` options feature works as expected. Does rely on the html-kit.com website to render the content for us. The alternative would be to run our own local http server that does the same thing. For now we can just live with the external service

PR #51, opened 2020-05-01
Viewport Override (for dynamic width/height) and Rspec tests compatibility with imagemagick7

inspiredstuffs:
Firstly, Thank you for this life saving gem!

I had to fork this gem for my personal use based on the following reasons:

- **Currently**, to use the viewport option you must pass both width and height parameter. But in my case, the required image width is known (for a Till printer) but the height varies.

**Changes made:**

- [x] Added a Javascript function to use the given viewport dimension (width or height) and then override the viewport dimension not specified with the document body offset dimension. 
- [x] Rspec tests for dynamic width and height scenarios
- [x] Rspec tests for compatibility with imagemagick7

I hope this will be useful for someone in need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ Gemfile.lock
 package-lock.json
 /node_modules
 /coverage
+spec/examples.txt
--- a/lib/grover.rb
+++ b/lib/grover.rb
@@ -86,7 +86,7 @@ class Grover
 
             // Setup viewport options (if provided)
             const viewport = options.viewport; delete options.viewport;
-            if (viewport != undefined) {
+            if (viewport != undefined && (viewport.width != undefined && viewport.height != undefined)) {
               await page.setViewport(viewport);
             }
 
@@ -123,6 +123,17 @@ class Grover
             const executeScript = options.executeScript; delete options.executeScript;
             if (executeScript != undefined) {
               await page.evaluate(executeScript);
+            }
+
+            // Override viewport with document body offsetHeight & offsetWidth, if viewport object is provided and the width or height isn't specified.
+            if (viewport != undefined && (viewport.width == undefined || viewport.height == undefined)) {
+              let vport = {};
+              vport['height'] = await page.evaluate(() => document.body.offsetHeight)
+              vport['width'] = await page.evaluate(() => document.body.offsetWidth)
+              if(viewport.width != undefined) { vport['width'] = viewport.width }
+              if(viewport.height != undefined) { vport['height'] = viewport.height }
+              const override = Object.assign(page.viewport(), vport);
+              await page.setViewport(override);
             }
 
             // If we're running puppeteer in headless mode, return the converted PDF
--- a/lib/grover.rb
+++ b/lib/grover.rb
@@ -84,9 +84,9 @@ class Grover
               request_options.timeout = timeout;
             }
 
-            // Setup viewport options (if provided)
+            // Setup viewport options (ONLY if width and height are provided)
             const viewport = options.viewport; delete options.viewport;
-            if (viewport != undefined && (viewport.width != undefined && viewport.height != undefined)) {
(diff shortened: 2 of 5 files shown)


abrom:
Thanks @inspiredstuffs 

A few things, assuming you're trying to capture an image (as apposed to a PDF), I would think you could achieve your desired outcome with:

```
options = {
  viewport: {
    width: <what ever you need to clamp it to>
    height: <some sort of sensible "minimum height">
  },
  full_page: true
}
```

The `full_page` option (see https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagescreenshotoptions) takes a screenshot of the full scrollable page. So as long as your page content only scrolls vertically, then happy days.

As such, I'm not seeing that your PR would be necessary.

Regarding your change however, a few points to note:
* Generally not a good idea to change the gem versions. Let the gem author manage that
* Generally a good idea to end all files new a newline (missing in the .gitignore)
* Your change doesn't handle cases where with viewport width or height is 0. Not sure that is a big issue, but something to consider (your change would NEVER set the viewport even when configured with 0 width or height)
* Please try abide by the line length rules applied to the code. Unfortunately Rubocop doesn't pick up line length foo for the heredoc, but the comments you have put in the grover code are huge!
* You seem to be mixing array and dot syntax for JS object assignment. Suggest stick with the most readable (ie dot syntax) `vport.width` etc :) See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects
* Try to stick with the general coding style. Specifically `if(...)` should be `if (...)` (missing space)
* You're assigning the width and height from the page evaluate unnecessarily. Better to just assign them once. ie 
```
vport.width = viewport.width ? viewport.width : await page.evaluate(() => document.body.offsetWidth)
```
* Try to use variable names that describe what they're doing. Not so sure that `vport` or `override` really do that
* Last, but probably most important.. I don't see how your code would actually fix the problem you've described. Assuming the content on your page will overflow by some means, assigning a fixed width after the initial render and using a fixed height from the initial render will invariably lead to content rendered off page (if the width becomes smaller) OR with your screenshot content with a bunch of whitespace at the bottom (if the width becomes larger). Neither sound very appealing, thus my suggestion to use the `full_page` option would likely work best ;)

inspiredstuffs:
Hello,

Thanks for your honest feedback. The aim of the PR was to share how I was able to adapt this gem to work for my scenario.

The viewport full page option didn’t work for my scenario.
This might explain my reason for doing this properly https://github.com/puppeteer/puppeteer/issues/2298

 Thanks again for the overall feedback

abrom:
Hmm not sure I follow. The puppeteer issue you linked suggests the problem was that the person was not specifying a height at all, rather than the problem being related to the `full_page` option!? If you specify the height option to something small, the page will render off screen (as expected) and the `full_page` option would then capture all of that off screen content. If i'm understanding your problem correctly, wouldn't that solve it?

abrom:
Taking that person's example and specifying a 'minimum' height:
```
const puppeteer = require('puppeteer');

(async() => {
    const browser = await puppeteer.launch();
    const page = await browser.newPage();
    await page.setViewport({width:500, height: 300}); //Custom Width
    await page.goto('https://reddit.com');
    await page.screenshot({path: 'example.png',fullPage:true});
    await browser.close();
})();
```
Results in a 500px wide and 4,000+ px high image. Isn't that what you're trying to achieve ? 

The same result would be achieved in Grover using the configs I specified above :)

inspiredstuffs:
Apologies for this unnecessary PR @abrom 

I am a complete idiot! 🤦

My desperation took me down this path. I tried the full_page option, but unknowingly with some other configurations that wasn't giving me the desired output. But your suggestion works perfect!

Thanks once again and good job 👍

abrom:
No worries @inspiredstuffs 👍 We all have those moments :) Just staring at it too long

PR #52, opened 2020-05-08
Update specs to support ImageMagick 7

abrom:
Update specs to support ImageMagick 7
Update Rubocop config to support latest cops
Include Puppeteer 3.x in CI build matrix
Bump minimum NodeJS version from CI default (8) to 10 - needed for Puppeteer 3.x


abrom:
FYI @inspiredstuffs I've pulled those ImageMagic 7 fixes you had proposed in #51. Thanks for that!

PR #57, opened 2020-05-21
Clarify cover page usage

### LINKED ISSUE #56: PDF Cover Pages are not rendered for non-middleware executions
Hi there, I appreciate the work you guys have done with this gem!

I'm not sure if this is intentional, but cover pages aren't rendered when you use direct execution,

```rb
# cover page will not render
Grover.new(pdf_report_url, front_cover_path: pdf_title_url).to_pdf
```

This is because the cover page logic exists in the middleware itself, but it is a little confusing because the grover instance receives the `front_cover_path` and `back_cover_path` options, it just doesn't do anything with them on its own.

Upon close examination of the README, the Cover Page does mention the middleware, but I figured the information about the global/meta-tag configurations was for the sake of the middleware execution, not that it was a complete restriction.

If the middleware restriction is intentional, I think it might help to point out the restriction explicitly. i.e. "This won't work for direct execution!"

As a fix, I used the same `combine_pdf` gem as the middleware uses:

```rb
  require 'combine_pdf'

  # ...

  def invoke(file_path)
    pdf = CombinePDF.parse(Grover.new(pdf_report_url).to_pdf)
    pdf >> CombinePDF.parse(Grover.new(pdf_title_url).to_pdf)
    pdf.save file_path
  end
```

This is an unrelated note and I don't expect you to try to resolve it, but when I tried to use middleware execution I would always get a Navigation Timeout error from the underlying puppeteer page.pdf() call. Running the same execution directly would quickly succeed. I don't have a good repro for that though (my app is a Single Page App that uses Hash Routing, which is why I ignored the middleware). I am assuming that the issue had to do with that app's architecture and isn't a middleware bug, but I figured I'd mention it in case that sparked an idea for you.

MrLeebo:
- Separate sections for Middleware vs Direct Execution
- Resolves #56

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,11 +254,14 @@ end
 Since the header/footer for Puppeteer is configured globally, displaying of front/back cover
 pages (with potentially different headers/footers etc) is not possible.
 
-To get around this, Grover's middleware allows you to specify relative paths for the cover page contents. For direct execution, you can make multiple calls and combine the resulting PDFs together.
+To get around this, Grover's middleware allows you to specify relative paths for the cover page contents.
+For direct execution, you can make multiple calls and combine the resulting PDFs together.
 
-**Using middleware**
+### Using middleware
 
-You can specify relative paths to the cover page contents using the `front_cover_path` and `back_cover_path` options either via the global configuration, or via meta tags. These paths (with query parameters) are then requested from the downstream app.
+You can specify relative paths to the cover page contents using the `front_cover_path` and `back_cover_path`
+options either via the global configuration, or via meta tags. These paths (with query parameters) are then
+requested from the downstream app.
 
 The cover pages are converted to PDF in isolation, and then combined together with the original PDF response,
 before being returned back up through the Rack stack.
@@ -284,9 +287,10 @@ Or via the meta tags in the original response:
 </html>
 ```
 
-**Direct execution
+### Direct execution
 
-To add a cover page using direct execution, you can make multiple calls and combine the results using the `combine_pdfs` gem.
+To add a cover page using direct execution, you can make multiple calls and combine the results using the
+`combine_pdfs` gem.
 
 ```rb
 require 'combine_pdf'
--- a/README.md
+++ b/README.md
@@ -250,12 +250,15 @@ end
 ```
 
 ## Cover pages
+
 Since the header/footer for Puppeteer is configured globally, displaying of front/back cover
 pages (with potentially different headers/footers etc) is not possible.
 
-To get around this, Grover's middleware allows you to specify relative paths for the cover page contents
-via `front_cover_path` and `back_cover_path` either via the global configuration, or via meta tags.
-These paths (with query parameters) are then requested from the downstream app.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,12 +250,15 @@ end
...
+
 Since the header/footer for Puppeteer is configured globally, displaying of front/back cover
 pages (with potentially different headers/footers etc) is not possible.
 
-To get around this, Grover's middleware allows you to specify relative paths for the cover page contents
-via `front_cover_path` and `back_cover_path` either via the global configuration, or via meta tags.
-These paths (with query parameters) are then requested from the downstream app.
+To get around this, Grover's middleware allows you to specify relative paths for the cover page contents. For direct execution, you can make multiple calls and combine the resulting PDFs together.
+
+**Using middleware**
```
abrom, line 259: Please use the markdown `###` notation to signify a 3rd level heading

```suggestion
### Using middleware
```

README.md
```diff
@@ -250,12 +250,15 @@ end
...
+
+You can specify relative paths to the cover page contents using the `front_cover_path` and `back_cover_path` options either via the global configuration, or via meta tags. These paths (with query parameters) are then requested from the downstream app.
```
abrom, line 261: Try to keep your line length < 100 chars to ease in reading in a text editor
```suggestion
You can specify relative paths to the cover page contents using the `front_cover_path` and `back_cover_path`
options either via the global configuration, or via meta tags. These paths (with query parameters) are then
requested from the downstream app.
```

README.md
```diff
@@ -281,6 +284,23 @@ Or via the meta tags in the original response:
 </html>
 ```
 
+**Direct execution
```
abrom, line 287:
```suggestion
### Direct execution
```

README.md
```diff
@@ -281,6 +284,23 @@ Or via the meta tags in the original response:
...
+
+To add a cover page using direct execution, you can make multiple calls and combine the results using the `combine_pdfs` gem.
```
abrom, line 289:
```suggestion
To add a cover page using direct execution, you can make multiple calls and combine the results using the
`combine_pdfs` gem.
```

README.md
```diff
@@ -250,12 +250,15 @@ end
...
 
 ## Cover pages
+
 Since the header/footer for Puppeteer is configured globally, displaying of front/back cover
 pages (with potentially different headers/footers etc) is not possible.
 
-To get around this, Grover's middleware allows you to specify relative paths for the cover page contents
-via `front_cover_path` and `back_cover_path` either via the global configuration, or via meta tags.
-These paths (with query parameters) are then requested from the downstream app.
+To get around this, Grover's middleware allows you to specify relative paths for the cover page contents. For direct execution, you can make multiple calls and combine the resulting PDFs together.
```
abrom, line 257:
```suggestion
To get around this, Grover's middleware allows you to specify relative paths for the cover page contents.
For direct execution, you can make multiple calls and combine the resulting PDFs together.
```

abrom:
Great, thanks @MrLeebo

PR #70, opened 2020-06-24
Passing cookies via middleware, ensure cookies are escaped

braindeaf:
It appears I missed a vital part of passing cookies from middleware, Rack::Utils unescapes cookies and they need to be re-escaped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## Unreleased
 
-### Added
-- (https://github.com/Studiosity/grover/pull/63) Ensure cookies from incoming request are passed to Grover via Middleware
-
-
 ## [0.12.2](releases/tag/v0.12.2) - 2020-06-23
 ### Fixed
 - [#66](https://github.com/Studiosity/grover/pull/66) Fix bug masking NodeJS launch failures
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Changelog
 
 ## Unreleased
-- None
+
+### Added
+- (https://github.com/Studiosity/grover/pull/63) Ensure cookies from incoming request are passed to Grover via Middleware
+
 
 ## [0.12.1](releases/tag/v0.12.1) - 2020-05-12
 ### Fixed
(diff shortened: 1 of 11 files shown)


spec/grover/middleware_spec.rb
```diff
@@ -383,11 +383,14 @@ def request_env
             with(
               'Grover McGroveryface',
               display_url: 'http://www.example.org/test',
-              cookies: [{ domain: 'www.example.org', name: 'key', value: 'value' }]
+              cookies: [
+                { domain: 'www.example.org', name: 'key', value: 'value' },
+                { domain: 'www.example.org', name: 'escaped', value: '%26%3D%3D'}
```
abrom, line 388: seems like a reasonable test, however i'd suggest it would also be good to have a test that actually calls to Grover/Puppeteer with some encoded cookies, have the page render the contents then check the result contains the expected encoded values. Hopefully that way there will be less risk of future regressions. This would probably live in the processor_spec

braindeaf: Agreed. I shall make it so.

braindeaf: Added that spec, but all it really does it prove that you can pass a cookies, escaped or not to Puppeteer, the real failure on this was not realising that the cookie parser had escaped it. My bad.

CHANGELOG.md
```diff
@@ -1,7 +1,6 @@
 # Changelog
 
 ## Unreleased
-- None
```
abrom, line 4: Can you please add an entry here for your change.  Likely
```
### Fixed
- [#70](https://github.com/Studiosity/grover/pull/70) Ensure cookies are escaped when passing cookies via middleware ([@braindeaf][])
```

spec/grover/processor_spec.rb
```diff
@@ -353,13 +353,15 @@
...
+                { 'name' => 'other-domain', 'value' => 'should not display', 'domain' => 'example.com' },
+                { 'name' => 'escaped', 'value' => '%26%3D%3D', 'domain' => 'cookie-renderer.herokuapp.com' }
               ]
             }
           end
 
-          it { expect(pdf_text_content).to include 'Request contained 1 cookie' }
+          it { expect(pdf_text_content).to include 'Request contained 2 cookies' }
           it { expect(pdf_text_content).to include '1. grover-test nom nom nom' }
+          it { expect(pdf_text_content).to include '2. escaped %26%3D%3D' }
```
abrom, line 364: Ok so this was actually a bug on my part on the cookie render app I wrote.. see https://github.com/abrom/cookie_renderer/commit/b724fb04d6a051886bab0e10ea12cfbb812b7cf0

With that fixed the spec now works how you would expect it to (ie decoded..)

```suggestion
          it { expect(pdf_text_content).to include '2. escaped &==' }
```

abrom:
FYI this has been released in v0.12.3

PR #71, opened 2020-07-05
Support addStyleTag and addScriptTag

paresharma:
This PR just supports for addStyleTag and addScriptTag which I needed for a recent project.

Would appreciate if this could be merged upstream and released as a patch or part of next release. Open to any feedback or changes needed.

### DIFF
--- a/lib/grover/js/processor.js
+++ b/lib/grover/js/processor.js
@@ -101,6 +101,18 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       }
     }
 
+    // add styles (if provided)
+    const addStyleTag = options.addStyleTag; delete options.addStyleTag;
+    if (addStyleTag !== undefined) {
+      await page.addStyleTag(addStyleTag);
+    }
+
+    // add scripts (if provided)
+    const addScriptTag = options.addScriptTag; delete options.addScriptTag;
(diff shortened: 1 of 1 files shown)


lib/grover/js/processor.js
```diff
@@ -101,6 +101,18 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       }
     }
 
+    // add styles (if provided)
+    const addStyleTag = options.addStyleTag; delete options.addStyleTag;
```
abrom, line 105: this should likely also be `styleTagOptions` (similarly for the script tag) given that's what they are..

abrom: it'd possibly also be an array (which would be enumerated to allow multiple style/script tag options)

paresharma: I used the names from the puppeteer doc, but agree to changing it to reflect what they really are. I'll add entries to README too.

> it'd possibly also be an array (which would be enumerated to allow multiple style/script tag options)

True, it is indeed an array. I'll update it.

abrom:
Any update on this @paresharma  ?

abrom:
Closing this due to inactivity. Feel free to re-open once you've addressed the feedback

Issue #81, opened 2020-09-22
JSON::ParserError: A JSON text must at least contain two octets!


ckhall:
```
"/Users/ckhall/Projects/project/vendor/bundle/ruby/2.6.0/gems/json-1.8.6/lib/json/common.rb:155:in `initialize'",
"/Users/ckhall/Projects/project/vendor/bundle/ruby/2.6.0/gems/json-1.8.6/lib/json/common.rb:155:in `new'",
"/Users/ckhall/Projects/project/vendor/bundle/ruby/2.6.0/gems/json-1.8.6/lib/json/common.rb:155:in `parse'",
"/Users/ckhall/Projects/project/vendor/bundle/ruby/2.6.0/gems/grover-0.13.1/lib/grover/processor.rb:84:in `call_js_method'",
"/Users/ckhall/Projects/project/vendor/bundle/ruby/2.6.0/gems/grover-0.13.1/lib/grover/processor.rb:20:in `convert'",
"/Users/ckhall/Projects/project/vendor/bundle/ruby/2.6.0/gems/grover-0.13.1/lib/grover.rb:50:in `to_pdf'",
"/Users/ckhall/Projects/project/app/services/create_order_labels_with_routing.rb:45:in `create_order_labels'",
```

did some digging and found the cause, at least as far as I can get anyways.  Line 84 of `lib/grover/processor.rb` where the value of `input` is a single newline `"\n"` cannot be parsed as json (obviously).

The size of the html string being passed to grover, with newlines and extraneous whitespace removed, is `3352813` bytes so I can only assume this may very well be the actual root cause.  The expected result is a little over 300 pages of labels.

I do not know what would cause a single newline to be returned on line 81 of `lib/grover/processor.rb`.  Smaller html strings are converted ok.

debug session:

```
    79: def call_js_method(method, url_or_html, options) # rubocop:disable Metrics/MethodLength
    80:   stdin.puts JSON.dump([method, url_or_html, options])
    81:   input = stdout.gets
    82:   raise Errno::EPIPE, "Can't read from worker" if input.nil?
    83: 
    84:   status, message, error_class = JSON.parse(input)
    85: 
    86:   if status == 'ok'
    87:     message
    88:   elsif error_class.nil?
    89:     raise Grover::JavaScript::UnknownError, message
    90:   else
    91:     raise Grover::JavaScript.const_get(error_class, false), message
    92:   end
    93: rescue JSON::ParserError => e
    94:   binding.pry
[...]

PR #82, opened 2020-09-23
rescue JSON::ParserError 

ckhall:
can happen when node craps out due to OOM error (issue #81)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 *.gem
 Gemfile.lock
 package-lock.json
+yarn.lock
+.bundle
 /node_modules
 /coverage
+/vendor
--- a/lib/grover/processor.rb
+++ b/lib/grover/processor.rb
@@ -90,7 +90,7 @@ class Grover
       else
         raise Grover::JavaScript.const_get(error_class, false), message
       end
-    rescue Errno::EPIPE, IOError
+    rescue Errno::EPIPE, IOError, JSON::ParserError
       raise Grover::Error, "Worker process failed:\n#{stderr.read}"
     end
 


ckhall:
#81

abrom:
Do you have any thoughts about how we can validate this with a spec test? Can we leverage what you found RE the fonts to cause this to fail? I'd definitely prefer to have some coverage of this to prevent future regressions.

ckhall:
@abrom so i looked through the specs to see if there were any tests against the existing exceptions that were being rescued and didn't see anything so I didn't add any tests.  This is a hard one for me.  Regardless of the specific issue I happened to experience, if `stdout.gets` returns something that causes a `JSON::ParserError` to be raised, I think it should be handled like any other exception that is already being rescued.  Could this simply be handled with stubs?

abrom:
Hi @ckhall you can see an example of the existing tests here: https://github.com/Studiosity/grover/blob/main/spec/grover/processor_spec.rb#L121-L134

The spec validates that a `nil` response from `stdout.gets` results in the Grover error being raised. I'm not a massive fan of this sort of spec as it does require in-depth knowledge of the inner workings of the thing being tested (in this case that the 2nd call to `gets` returning a `nil` value simulating a failure). Ideally we'd be black-box testing it. Having said that, some coverage is always going to be better than none.

You could possibly just return `"\n"` from the `stdout.gets` stub which would match your findings, but having a "real" test would certainly be more effective at ensuring this fix stays the most relevant. Could you mock up some html of similar size to what you were generating and have that cause the failure? 

```
<<~HTML
<html>
  <head>
    ... include a reference to an external font? Google hosts plenty, hopefully those replicate this issue?
    <style>
      * {
         ... use the font
      }
  <head>

  <body>
    #{Array.new(38000) { '<div>Some content that if displayed enough times will cause memory overflow issues</div>' }
  </body>
</html>
HTML
```

38,000 * 88 bytes approximates the document you were testing..

ckhall:
@abrom so just an update, i started down this path and for the life of me, I cannot get anything to cause an OOM error, even generating an output pdf > 400MB.

I have to wonder if it's possibly related to the fonts themselves and/or the particular html and css being used.  Unfortunately the fonts we are using are not FOSS.  The fonts are `Acta-Book` and `SuperGroteskA-BdLF`, both `.otf`fonts (not sure if that matters.

When I find some time, I'll try and replicate the html and css and see if I can duplicate with similar fonts.

abrom:
Hmm weird one.

Don't spend too much time on it. If it's not going to work then a stubbed option would be ok.

Thank you for trying though!

abrom:
Any progress on this @ckhall ?

ckhall:
@abrom sorry, had life catch up to me for a while.  I was not able to duplicate it.  Unfortunately, the code that was causing the issue is not open source and I no longer work for the company.  My apologies for dropping off for a bit there.

abrom:
No worries @ckhall

I wasn't set on the full reproduction and the stub is likely fine.

Thanks for the PR!

PR #84, opened 2020-10-01
Add root_url option to middleware

### LINKED ISSUE #83: Can't override display_url when using middleware
If I'm reading this correctly, there is no way to override the detected display_url option when using the middleware.
https://github.com/Studiosity/grover/blob/main/lib/grover/middleware.rb#L104

I need to do this because the Rails app is running in a docker container, so the url that Grover needs to use is different from the one I'm using to access from the host.

anamba:
Fixes #83 

Oh my, this test suite is... not fast. But that's understandable. It took a while to figure out where to add the tests, but I think I found a good place. Let me know what you think.

### DIFF
--- a/lib/grover/middleware.rb
+++ b/lib/grover/middleware.rb
@@ -16,9 +16,7 @@ class Grover
       @png_request = false
       @jpeg_request = false
 
-      options = Grover.configuration.options
-      @root_url = options[:root_url]
-      @root_url = args.last[:root_url] if args.last.is_a? Hash
+      @root_url = args.last.is_a?(Hash) ? args.last[:root_url] : Grover.configuration.options[:root_url]
     end
 
     def call(env)
--- a/lib/grover/middleware.rb
+++ b/lib/grover/middleware.rb
@@ -10,11 +10,15 @@ class Grover
   # @see https://github.com/pdfkit/pdfkit
   #
   class Middleware # rubocop:disable Metrics/ClassLength
-    def initialize(app)
+    def initialize(app, *args)
       @app = app
       @pdf_request = false
       @png_request = false
       @jpeg_request = false
+
+      options = Grover.configuration.options
+      @root_url = options[:root_url]
+      @root_url = args.last[:root_url] if args.last.is_a? Hash
     end
 
     def call(env)
--- a/spec/grover/middleware_spec.rb
+++ b/spec/grover/middleware_spec.rb
@@ -431,6 +431,50 @@ describe Grover::Middleware do
         get 'http://www.example.org/test.pdf'
         expect(last_response.body.bytesize).to eq Grover.new('Processed McProcessyface').to_pdf.bytesize
       end
+
+      context 'with root_url specified' do
+        subject(:mock_app) do
+          builder = Rack::Builder.new
+          builder.use described_class, root_url: 'http://example.com/'
+          builder.run downstream
+          builder.to_app
+        end
+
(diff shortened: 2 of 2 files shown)


lib/grover/middleware.rb
```diff
@@ -10,11 +10,15 @@ class Grover
...
-    def initialize(app)
+    def initialize(app, *args)
       @app = app
       @pdf_request = false
       @png_request = false
       @jpeg_request = false
+
+      options = Grover.configuration.options
+      @root_url = options[:root_url]
+      @root_url = args.last[:root_url] if args.last.is_a? Hash
```
abrom, line 21: could simplify this with:
```
@root_url = args.last.is_a?(Hash) ? args.last[:root_url] : Grover.configuration.options[:root_url]
```

spec/grover/middleware_spec.rb
```diff
@@ -431,6 +431,50 @@
...
+            receive(:process).
+              with('Grover McGroveryface', 'http://example.com/', 'http').
+              and_return('Processed McProcessyface')
+          )
+          get 'http://www.example.org/test.pdf'
+          expect(last_response.body.bytesize).to eq Grover.new('Processed McProcessyface').to_pdf.bytesize
+        end
+      end
+
+      context 'converts relative paths' do
```
abrom, line 454: Rubocop is complaining because it wants the contexts to start with `with`, `without` or `when`.

```
context 'when the response contains relative paths'
```
maybe?

spec/grover/middleware_spec.rb
```diff
@@ -431,6 +431,50 @@
...
+            get 'http://www.example.org/test.pdf'
+            expect(last_response.body.bytesize).to eq Grover.new('src="http://example.com/asdf"').to_pdf.bytesize
+          end
+        end
+
+        context 'without root_url specified' do
+          it 'uses the detected root_url (request url)' do
+            get 'http://www.example.org/test.pdf'
+            expect(last_response.body.bytesize).to eq Grover.new('src="http://www.example.org/asdf"').to_pdf.bytesize
+          end
```
abrom, line 475: it would also be nice to see a spec (or two) test your code fetching the `root_url` from the configuration options. One to test it fetches from there, and a second to show it's overloaded by the initialiser option

abrom (changes requested):
Change is looking good @anamba but a few things I think could be better.

Can you please take a look, and let me know if you have any questions 👍

anamba:
Finally switching back to the project that required this gem...

abrom:
No worries @anamba. I see the build is still failing but that looks to have been an issue with how I had setup the call to Rubocop for Travis. I've fixed that in #86 but note that I've also addressed a new cop aimed at preventing `and_return` with expectations. ie:
bad:
```
expect(foo).to receive(:bar).and_return 'baz'
```
good:
```
allow(foo).to receive(:bar).and_return 'baz'
expect(foo).to receive(:bar)
```
Could you please rebase your branch off main (will fix the build issue), and fix the rubocop lint that is going to come from:
https://github.com/Studiosity/grover/pull/84/files#diff-b9c6602f53d1c3a53626aae8906d58c842e1b0a70ab2dab7bb71face3bdee978R444-R448

You can see plenty of examples of how I've addressed the lint issue in #86

Apologies for the run-about..

abrom:
Any progress on this @anamba ?

anamba:
The project I was using Grover for is on hold indefinitely, so I'm not sure when I'll have time to take another look at this.

PR #90, opened 2020-11-18
Update gem name for combining PDF files

xiazek:
`combine_pdf` seems to be a proper [gem name](https://github.com/boazsegev/combine_pdf)

```
➜   gem install combine_pdfs
ERROR:  Could not find a valid gem 'combine_pdfs' (>= 0) in any repository
ERROR:  Possible alternatives: combine_pdf, compile_pdf
```


abrom:
Oops.. good catch @xiazek thanks

PR #95, opened 2020-12-07
Handle `nil` and empty string content edge-cases

### LINKED ISSUE #94: Call Grover.new('').to_pdf gets stuck forever
Happy path:
  - When you call `to_pdf` with a value with content (even spaces), it returns the PDF

Bad path:
  - When you call `to_pdf` with a nil value, it returns an error. I don't know if it should return an error or change it to a blank value and return the PDF.
  - When you call `to_pdf` with an empty value (`Grover.new('').to_pdf`), it gets frozen until it reaches the timeout.

abrom:
Resolves #94 

Handle `nil` input case by treating it as an empty string (convert all input to a string with `to_s`)

Handle empty string case in the processor by passing a single space to Puppeteer instead. Likely cause is that Puppeteer is taking an empty string to mean "not provided" and thus ignoring the call to `respond`, instead choosing to wait for further input.. which never comes:

https://github.com/puppeteer/puppeteer/blob/v5.5.0/docs/api.md#httprequestrespondresponse

### DIFF
--- a/lib/grover.rb
+++ b/lib/grover.rb
@@ -33,8 +33,8 @@ class Grover
   #   see https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagepdfoptions
   #
   def initialize(url, options = {})
-    @url = url
-    @options = OptionsBuilder.new(options, url)
+    @url = url.to_s
+    @options = OptionsBuilder.new(options, @url)
     @root_path = @options.delete 'root_path'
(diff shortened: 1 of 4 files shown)


abrom:
@alexwbuschle how does this look to you? It also means that Grover would "deal" with other input types too.. ie if someone passed in some other type than a String (or `NilClass`)

alexwbuschle (approved):
👏

PR #107, opened 2021-02-20
Add support to return the html content

paresharma:
This PR adds support to retrieve the HTML content. I just find it particularly helpful in debugging PDFs.
Let me know what you think about this 🙌 

FYI: https://github.com/puppeteer/puppeteer/pull/419

### DIFF
--- a/spec/grover/processor_spec.rb
+++ b/spec/grover/processor_spec.rb
@@ -581,5 +581,22 @@ describe Grover::Processor do
         colours.map { |colour| image.data.dig('channelStatistics', colour, 'mean').to_s }
       end
     end
+
+    context 'when rendering HTML' do
+      let(:method) { :content }
+
+      let(:url_or_html) do
+        <<-HTML
+          <html>
+            <head></head>
+            <body>
+              <h1>Hey there</h1>
+            </body>
+          </html>
+        HTML
+      end
+
+      it { expect(convert.gsub(/[[:space:]]+/, '')).to eq url_or_html.gsub(/[[:space:]]+/, '') }
+    end
   end
 end
--- a/spec/grover/processor_spec.rb
+++ b/spec/grover/processor_spec.rb
@@ -591,12 +591,30 @@ describe Grover::Processor do
             <head></head>
             <body>
               <h1>Hey there</h1>
+              <h2>Konnichiwa</h2>
             </body>
           </html>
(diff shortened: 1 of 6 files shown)


spec/grover/processor_spec.rb
```diff
@@ -581,5 +581,22 @@ def mean_colour_statistics(image)
...
+          <html>
+            <head></head>
+            <body>
+              <h1>Hey there</h1>
+            </body>
+          </html>
+        HTML
+      end
+
+      it { expect(convert.gsub(/[[:space:]]+/, '')).to eq url_or_html.gsub(/[[:space:]]+/, '') }
```
abrom, line 599: Two things here.. I'd suggest using `Grover::Utils.squish` rather than removing all spaces. Would still give a consistent result whilst being true to the rendered output. Also, it feels like this would be a good opportunity to test that the *rendered* HTML is returned. ie something like: 
```
    context 'when rendering HTML' do
      let(:method) { :content }

      let(:url_or_html) do
        <<-HTML
          <html>
            <head></head>
            <body>
              <h1>Hey there</h1>
              <h2>Konnichiwa</h2>
              <script>document.querySelector('h1').remove();</script>
            </body>
          </html>
        HTML
      end

      it 'returns the rendered HTML' do
        expect(Grover::Utils.squish(convert)).to eq Grover::Utils.squish(<<~HTML)
          <html><head></head>
            <body>
              <h2>Konnichiwa</h2>
              <script>document.querySelector('h1').remove();</script>
            </body></html>
        HTML
      end
    end
```
Ie not just what you gave it, but rather the result of it being rendered.

abrom (changes requested):
Looks like a good change @paresharma 👍 

I've got one suggestion about the processor specs if you can take a look

spec/grover/processor_spec.rb
paresharma: Good catch and point, thanks. Updated 🙌

abrom (approved):
Nice one @paresharma

paresharma:
Great thanks. I would appreciate if you could release these changes 🙌

abrom:
Released in [0.14.2](https://rubygems.org/gems/grover/versions/0.14.2)

paresharma:
Thank you 🙌

PR #112, opened 2021-05-27
Feature: Add flag to fail to generate PDF if assets fail to load.

deanmarano:
Hi there! Grover has made our lives so much easier, thank you!

I'm hoping to add in a feature to allow PDF generation to fail when an asset fails to load. I've added a test and a line to the README. Any feedback or help getting this in would be greatly appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ The `wait_for_selector` option can also be used to wait until an element appears
 
 The `wait_for_timeout` option can also be used to wait the specified number of milliseconds have elapsed.
 
-The `request_failure` option, when enabled, will raise a `Grover::JavaScript::RequestFailedError`
+The `raise_on_request_failure` option, when enabled, will raise a `Grover::JavaScript::RequestFailedError`
 if the initial content request or any subsequent asset request returns a bad response or times out.
 
 The Chrome/Chromium executable path can be overridden with the `executable_path` option.
--- a/README.md
+++ b/README.md
@@ -154,7 +154,11 @@ The `wait_for_selector` option can also be used to wait until an element appears
 
 The `wait_for_timeout` option can also be used to wait the specified number of milliseconds have elapsed.
 
+The `request_failure` option can be used to fail generation if assets fail to load.
+
 The Chrome/Chromium executable path can be overridden with the `executable_path` option.
+
+
 
 Javascript can be executed on the page (after render and before conversion to PDF/image)
 with the `execute_script` option.
--- a/README.md
+++ b/README.md
@@ -154,7 +154,8 @@ The `wait_for_selector` option can also be used to wait until an element appears
 
 The `wait_for_timeout` option can also be used to wait the specified number of milliseconds have elapsed.
 
-The `request_failure` option can be used to fail generation if assets fail to load.
+The `request_failure` option, when enabled, will raise a `Grover::JavaScript::RequestFailedError`
+if the initial content request or any subsequent asset request returns a bad response or times out.
 
 The Chrome/Chromium executable path can be overridden with the `executable_path` option.
 
--- a/README.md
+++ b/README.md
@@ -158,8 +158,6 @@ The `request_failure` option can be used to fail generation if assets fail to lo
 
 The Chrome/Chromium executable path can be overridden with the `executable_path` option.
 
-
-
 Javascript can be executed on the page (after render and before conversion to PDF/image)
 with the `execute_script` option.
 
--- a/lib/grover/js/processor.js
+++ b/lib/grover/js/processor.js
@@ -125,9 +125,9 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
     }
 
     const failOnRequest = options.requestFailure; delete options.requestFailure;
-    if (failOnRequest !== undefined) {
+    if (failOnRequest) {
       page.on('requestfinished', (request) => {
-        if(request.response() && !request.response.ok()) {
+        if (request.response() && !request.response.ok()) {
           errors.push(request);
         }
       });
@@ -203,10 +203,10 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
     }
 
     if (errors.length > 0) {
-      function RequestFailedError(errors = []) {
+      function RequestFailedError(errors) {
         this.name = "RequestFailedError";
         this.message = errors.map(e => {
-          if(e.failure()) {
+          if (e.failure()) {
             return e.failure().errorText + " at " + e.url();
           } else if (e.response() && e.response().status()) {
             return e.response.status() + " " + e.url();
--- a/lib/grover/js/processor.js
+++ b/lib/grover/js/processor.js
@@ -124,10 +124,10 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       page.setGeolocation(geolocation);
     }
 
-    const failOnRequest = options.requestFailure; delete options.requestFailure;
-    if (failOnRequest) {
+    const raiseOnRequestFailure = options.raiseOnRequestFailure; delete options.raiseOnRequestFailure;
+    if (raiseOnRequestFailure) {
       page.on('requestfinished', (request) => {
-        if (request.response() && !request.response.ok()) {
+        if (request.response() && !request.response().ok() && !request.redirectChain()) {
           errors.push(request);
         }
       });
@@ -209,7 +209,7 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
           if (e.failure()) {
             return e.failure().errorText + " at " + e.url();
           } else if (e.response() && e.response().status()) {
-            return e.response.status() + " " + e.url();
+            return e.response().status() + " " + e.url();
           } else {
             return "UnknownError " + e.url()
           }
--- a/lib/grover/js/processor.js
+++ b/lib/grover/js/processor.js
@@ -127,7 +127,7 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
     const raiseOnRequestFailure = options.raiseOnRequestFailure; delete options.raiseOnRequestFailure;
     if (raiseOnRequestFailure) {
       page.on('requestfinished', (request) => {
-        if (request.response() && !request.response().ok() && !request.redirectChain()) {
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -154,8 +154,12 @@ The `wait_for_selector` option can also be used to wait until an element appears
 
 The `wait_for_timeout` option can also be used to wait the specified number of milliseconds have elapsed.
 
+The `request_failure` option can be used to fail generation if assets fail to load.
+
 The Chrome/Chromium executable path can be overridden with the `executable_path` option.
 
+
+
```
deanmarano, line 162:
```suggestion
```

spec/grover/processor_spec.rb
```diff
@@ -490,6 +490,24 @@
...
+        let(:url_or_html) do
+          <<-HTML
+            <html>
+              <body>
+                <img src="http://foo.bar/baz.img" />
+              </body>
+            </html>
+          HTML
+        end
+        let(:options) { basic_header_footer_options.merge('requestFailure' => true) }
```
deanmarano, line 503:
```suggestion
        let(:options) { basic_header_footer_options.merge('requestFailure' => true) }

```

lib/grover/js/processor.js
```diff
@@ -124,6 +124,18 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       page.setGeolocation(geolocation);
     }
 
+    const failOnRequest = options.requestFailure; delete options.requestFailure;
+    if (failOnRequest !== undefined) {
```
abrom, line 128: this would still enable the error reporting if `requestFailure: ''` or `requestFailure: false` which would not appear to be the intent. Suggest:
```suggestion
    if (failOnRequest) {
```

lib/grover/js/processor.js
```diff
@@ -124,6 +124,18 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
...
+      page.on('requestfinished', (request) => {
+        if(request.response() && !request.response.ok()) {
```
abrom, line 130: Prefer `if (`. Sorry, I haven't setup style linting for the JS code. Will need to add it to my list :)
```suggestion
        if (request.response() && !request.response.ok()) {
```

lib/grover/js/processor.js
```diff
@@ -190,6 +202,23 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       await page.hover(hoverSelector);
     }
 
+    if (errors.length > 0) {
+      function RequestFailedError(errors = []) {
+        this.name = "RequestFailedError";
+        this.message = errors.map(e => {
+          if(e.failure()) {
```
abrom, line 209: same here (`if (`)
```suggestion
          if (e.failure()) {
```

lib/grover/js/processor.js
```diff
@@ -190,6 +202,23 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
...
+            return e.failure().errorText + " at " + e.url();
+          } else if (e.response() && e.response().status()) {
+            return e.response.status() + " " + e.url();
+          } else {
+            return "ERR" + " " + e.url()
```
abrom, line 214: Two things here. 
1. "ERR" seems maybe a tad vague. It seems like timeouts etc would be handled by the `e.failure()` case and bad responses by the `e.response().status()` case. What do you see this handling? If it's "unknown" then suggest using language to that point.
2. `"ERR " + e.url()` would seem simpler (albeit after addressing point 1.)

lib/grover/js/processor.js
```diff
@@ -190,6 +202,23 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       await page.hover(hoverSelector);
     }
 
+    if (errors.length > 0) {
+      function RequestFailedError(errors = []) {
```
abrom, line 206: do you need a default here given we know there are errors from the line above?

README.md
```diff
@@ -154,6 +154,8 @@ The `wait_for_selector` option can also be used to wait until an element appears
 
 The `wait_for_timeout` option can also be used to wait the specified number of milliseconds have elapsed.
 
+The `request_failure` option can be used to fail generation if assets fail to load.
```
abrom, line 157: Might be worth expanding here in terms of what is expected when enabled. Something maybe more along the lines of:
```suggestion
The `request_failure` option, when enabled, will raise a `Grover::JavaScript::RequestFailedError`
if the initial content request or any subsequent asset request returns a bad response or times out.
```

abrom (changes requested):
Hi @deanmarano the idea is solid and the change is looking great. I've left a few comments but nothing too drastic. Take a look and let me know what you think. 👍

lib/grover/js/processor.js
deanmarano: No, just being over-defensive :-P Will remove.
```suggestion
      function RequestFailedError(errors) {
```

lib/grover/js/processor.js
deanmarano: I put this in before the code was properly handling the previous cases. I think the way the code is structured, I would prefer to keep a `else` condition (so we don't have things in the `errors` array that don't have error messages) so I'll change this to state it's an unknown error.

lib/grover/js/processor.js
```diff
@@ -124,6 +124,18 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       page.setGeolocation(geolocation);
     }
 
+    const failOnRequest = options.requestFailure; delete options.requestFailure;
+    if (failOnRequest) {
+      page.on('requestfinished', (request) => {
+        if (request.response() && !request.response.ok()) {
```
abrom, line 130: hmm looks like this should be `!request.response().ok()`  ?

```suggestion
        if (request.response() && !request.response().ok()) {
```

lib/grover/js/processor.js
```diff
@@ -190,6 +202,23 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
...
     }
 
+    if (errors.length > 0) {
+      function RequestFailedError(errors) {
+        this.name = "RequestFailedError";
+        this.message = errors.map(e => {
+          if (e.failure()) {
+            return e.failure().errorText + " at " + e.url();
+          } else if (e.response() && e.response().status()) {
+            return e.response.status() + " " + e.url();
```
abrom, line 212: and shouldn't this be `e.response().status()` ?
```suggestion
            return e.response().status() + " " + e.url();
```
I think this also deserves a separate spec test. Ie one where you include a URI to a host that DOES exist, but where the image path doesn't (ie get a 404 back and thus test this part of the code). Say http://google.com/foo.png

spec/grover/processor_spec.rb
```diff
@@ -490,6 +490,25 @@
         it { expect(pdf_text_content).to eq "#{date} Hey there http://www.example.net/foo/bar 1/1" }
       end
 
+      context 'when request failure option is specified' do
+        let(:url_or_html) do
```
abrom, line 494: per above, this likely should be two separate contexts under the "request failure option specified" context (bad host VS bad path) 

I'd also suggest testing an external source that DOES work, say https://placekitten.com/200/200 (ie test that with valid image sources that it doesn't raise an error, but instead returns a PDF.

README.md
```diff
@@ -154,6 +154,9 @@ The `wait_for_selector` option can also be used to wait until an element appears
 
 The `wait_for_timeout` option can also be used to wait the specified number of milliseconds have elapsed.
 
+The `request_failure` option, when enabled, will raise a `Grover::JavaScript::RequestFailedError`
```
abrom, line 157: hmm the more I think about it the more this option name doesn't really fit. How about `raise_request_failures` or `raise_on_request_failure` ? Just tells the person reading the config exactly what to expect when enabled. Cause you know.. no one actually WANTS a request failure right?? ie `request_failure: true` would seem to suggest it will always fail :)

abrom (changes requested):
Getting really close @deanmarano

I've noticed a few bugs as well as a couple of missing specs (ie test the success result when the option is enabled, but no errors.. and test for a 404).

I've also left a comment about the option name itself. I think as it currently stands it's a bit vague about what it'll do when enabled. A slightly more verbose option name should get rid of any of that ambiguity.

README.md
deanmarano: I like `raise_on_request_failure`. I was trying to match the puppeteer function naming, but I think that your suggestion is better.

lib/grover/js/processor.js
deanmarano: Added some more tests (one success case, one redirect) to expose this.

spec/grover/processor_spec.rb
deanmarano: Agreed, done.

lib/grover/js/processor.js
deanmarano: You are correct.

spec/grover/processor_spec.rb
```diff
@@ -490,6 +490,50 @@
...
+                  <img src="http://foo.bar/baz.img" />
+                </body>
+              </html>
+            HTML
+          end
+
+          it do
+            expect do
+              convert
+            end.to raise_error Grover::JavaScript::RequestFailedError, %r{net::ERR_NAME_NOT_RESOLVED at http://foo.bar/baz.img}
```
abrom, line 511: Hmm you seem to have missed the test case for a 404? Ie valid host, invalid path.

abrom (changes requested):
Hi @deanmarano it looks like there's still one test case missing (404). Can you please add that in? Suggest something like https://google.com/invalid.png or similar

spec/grover/processor_spec.rb
deanmarano: You're right! Thanks for working through this with me and turning this into a real PR. 😎

deanmarano:
@abrom It looks like an expired w3.org cert is causing some tests to fail. I would guess they'll fix it relatively quickly?

abrom (approved):
Looks great @deanmarano 👍 

Thanks for the contribution!

abrom:
FYI this has been released in [v1.0.1](https://rubygems.org/gems/grover/versions/1.0.1)

PR #117, opened 2021-07-19
Add waitForFunction support

rafraser:
This PR adds support for waitForFunction & some relevant options.

My use case for this is when printing a page with Mapbox, it's good to wait until the map has loaded - there's not really a great way to handle this with the current wait variants. 

eg:
`page.waitForFunction("map.loaded()")`

I've added a couple of rspec tests for this functionality, and tested it properly on some pages with maps as well. Let me know if you have any thoughts!

### DIFF
--- a/spec/grover/processor_spec.rb
+++ b/spec/grover/processor_spec.rb
@@ -490,6 +490,30 @@ describe Grover::Processor do
         it { expect(pdf_text_content).to eq "#{date} Hey there http://www.example.net/foo/bar 1/1" }
       end
 
+      context 'when wait for function option is specified' do
+        let(:url_or_html) do
+          <<-HTML
+            <html>
+              <body></body>
+
+              <script>
+                setTimeout(function() {
+                  document.body.innerHTML = '<h1 id="test">Hey there</h1>';
+                }, 100);
+              </script>
+            </html>
+          HTML
+        end
+        let(:options) do
+          basic_header_footer_options.merge(
+            'waitForFunction' => 'document.getElementById("test") !== null'
+          )
+        end
+        let(:date) { Date.today.strftime '%-m/%-d/%Y' }
+
+        it { expect(pdf_text_content).to eq "#{date} Hey there http://www.example.net/foo/bar 1/1" }
+      end
+
       context 'when raise on request failure option is specified' do
         let(:options) { basic_header_footer_options.merge('raiseOnRequestFailure' => true) }
         let(:date) { Date.today.strftime '%-m/%-d/%Y' }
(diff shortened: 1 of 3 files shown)


rafraser:
looks like codeclimate is yelling at my tests being too similar to the waitForSelector ones - anything I can do about that?

spec/grover/processor_spec.rb
```diff
@@ -579,6 +603,46 @@
         it { expect(pdf_text_content).to eq "#{date} http://www.example.net/foo/bar 1/1" }
       end
 
+      context 'when wait for function option is specified with options' do
```
abrom, line 606: Any reason this doesn't live up under the other test you added? Would be good to keep the two together.

spec/grover/processor_spec.rb
```diff
@@ -579,6 +603,46 @@
...
+            basic_header_footer_options.merge(
+              'waitForFunction' => 'document.getElementById("test") !== null',
+              'waitForFunctionOptions' => { "polling": 50, "timeout": 250 }
+            )
+          end
+
+          it 'raises a JavaScript error if waitForFunction fails' do
+            expect do
+              pdf_text_content
+            end.to raise_error Grover::JavaScript::Error
```
abrom, line 641: suggest also adding either a string match for the message, or if that gets complicated, a regex match. ie: 
```
end.to raise_error Grover::JavaScript::Error, 'The full error message'
```
or

```
end.to raise_error Grover::JavaScript::Error, /partial match/
```
I would expect the match to be something about it being a timeout. Just makes following the intent of the test a little clearer

abrom (changes requested):
Looks great @rafraser 👍 I've had adding this on my list for a while and just hadn't got around to it

I've left a few suggestions just to tidy things up. Can you please take a look

abrom:
> looks like codeclimate is yelling at my tests being too similar to the waitForSelector ones - anything I can do about that?

No worries.. to be expected with those sorts of things 👍

abrom:
Btw looks like that timeout spec of yours is failing:

```
  1) Grover::Processor#convert when converting to PDF when wait for function option is specified with options when waiting for function takes too long raises a JavaScript error if waitForFunction fails
     Failure/Error:
       expect do
         pdf_text_content
       end.to raise_error Grover::JavaScript::Error
     
       expected Grover::JavaScript::Error but nothing was raised
     # ./spec/grover/processor_spec.rb:639:in `block (6 levels) in <top (required)>'
```

I haven't looked too deeply into why, but let me know if you need any help figuring it out

spec/grover/processor_spec.rb
```diff
@@ -579,6 +603,46 @@
...
+        end
+        let(:date) { Date.today.strftime '%-m/%-d/%Y' }
+
+        it { expect(pdf_text_content).to eq "#{date} Hey there http://www.example.net/foo/bar 1/1" }
+
+        context 'when waiting for function takes too long' do
+          let(:options) do
+            basic_header_footer_options.merge(
+              'waitForFunction' => 'document.getElementById("test") !== null',
+              'waitForFunctionOptions' => { "polling": 50, "timeout": 250 }
```
abrom, line 634: btw I would suggest as a means to avoid duplicating the setup, and to making it clearer what exactly is different about this content.. instead put the timeout value into a let variable up above, then this would just become: 
```
let(:wait_function_timeout) { 250 }
```

(and the bit up above being)
```
let(:options) do
  basic_header_footer_options.merge(
    'waitForFunction' => 'document.getElementById("test") !== null',
    'waitForFunctionOptions' => { "polling": 50, "timeout": wait_function_timeout }
  )
end
let(:wait_function_timeout) { 750 }
``

spec/grover/processor_spec.rb
rafraser: My logic here was that I put:
 `when wait for function option is specified` right under `when wait for selector option is specified`
 and similarly:
 `... is specified with options` right under `when wait for selector option is specified with options` 
 
 But I do agree that it makes more sense to keep them close together instead

spec/grover/processor_spec.rb
rafraser: Updated this to be `Grover::JavaScript::TimeoutError, /waiting for function failed/`

rafraser:
Thanks for the feedback! I've tidied up the spec as per your suggestions & refactored the JS so it's a truer reflection of 'waiting for a function'. 

Should also fix that failing timeout spec - previously Puppeteer sometimes took long enough to spin up that the script had already wrapped up by the time waitForFunction was being called.

abrom (approved):
Looks great @rafraser 👍 

Thanks for the change

abrom:
Released in [v1.0.2](https://github.com/Studiosity/grover/releases/tag/v1.0.2)

PR #122, opened 2021-08-19
Fix #121: stylesheets/images rendering with puppeteer v10.2.0

### LINKED ISSUE #121: Upgrading to puppeteer 10.2.0 breaks stylesheet rendering
Assets / Stylesheets aren't fetched after upgrading to Puppeteer 10.2.0. Puppeteer 10.1.0 works fine.

mkalygin:
In puppeteer v10.2.0 they introduced changes to the way request interception handlers are processed:

- https://github.com/puppeteer/puppeteer/compare/v10.1.0...v10.2.0
- https://github.com/puppeteer/puppeteer/blob/c510df8d8e52fa361c26f6db011aaa9410c8996a/src/common/Page.ts#L627-L633

It caused a race condition in processing between two handlers (`page.once` and `page.on`). So I fixed this problem by properly awaiting response and disabling interception.

### DIFF
--- a/lib/grover/js/processor.js
+++ b/lib/grover/js/processor.js
@@ -145,11 +145,15 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       // Request is some HTML content. Use request interception to assign the body
       requestOptions.waitUntil = waitUntil || 'networkidle0';
       await page.setRequestInterception(true);
-      page.once('request', async request => {
-        await request.respond({ body: urlOrHtml === '' ? ' ' : urlOrHtml });
-        // Reset the request interception
-        // (we only want to intercept the first request - ie our HTML)
-        await page.setRequestInterception(false);
+      let htmlIntercepted = false;
+      page.on('request', request => {
+        // We only want to intercept the first request - ie our HTML
+        if (htmlIntercepted)
+          request.continue();
+        else {
+          htmlIntercepted = true
+          request.respond({ body: urlOrHtml === '' ? ' ' : urlOrHtml });
+        }
       });
       const displayUrl = options.displayUrl; delete options.displayUrl;
       await page.goto(displayUrl || 'http://example.com', requestOptions);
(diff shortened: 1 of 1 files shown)


abrom:
Good find @mkalygin 👍 

Although I did see that Puppeteer will occasionally error out using this solution with:
```
Error: Request Interception is not enabled!
```

What did appear to work ok was:
```
      await page.setRequestInterception(true);
      let htmlIntercepted = false;
      page.on('request', request => {
        // We only want to intercept the first request - ie our HTML
        if (htmlIntercepted)
          request.continue();
        else {
          htmlIntercepted = true
          request.respond({ body: urlOrHtml === '' ? ' ' : urlOrHtml });
        }
      });
```
Thoughts?

mkalygin:
@abrom I'm wondering how is it even possible, given that I use two `await` calls and they're guaranteed to be synchronous.

Are you saying that `await request.respond` isn't guaranteed to be fully processed?

Is there any way to repro it? Or are you referring to some previous observations of that issue?

abrom:
Good question! But I don't have an answer. I'm thinking it's a timing thing where the browser is going ahead and starting the next request BEFORE the request interception has been disabled. Maybe the error message is just a bit confusing?

In any case, easy enough to test it:

```
npm install puppeteer@10.2.0
bundle exec rspec spec/grover/processor_spec.rb
```

mkalygin:
@abrom interesting, I can't reproduce the issue you mention on my local machine. The CI is all green too.

With your implementation the example code I provided in #121 doesn't work - it just hangs and quit execution by timeout.

abrom:
I would note that the CI was only passing because it was set to test up to v10.1.0. I pushed an update in #123 to include v10.2.0 yesterday so if you rebase that should give a better idea of what the state of the issue is.

Something tells me we haven't got to the bottom of this if your fix crashes for me and mine hangs for you!

abrom:
By the way, did you copy my entire snippet ? Note that it needs to be `page.on` instead of `page.once` (I expect that would cause the hanging issue you mentioned..)

mkalygin:
@abrom you're absolutely right, I didn't notice this change. :–) Updated the PR.

abrom (approved):
Nice one @mkalygin thanks!

abrom:
FYI released in [v1.0.3](https://rubygems.org/gems/grover/versions/1.0.3)

Issue #127, opened 2021-08-28
Error when using vision_deficiency


julianwegkamp:
Hi,

When setting vision_deficiency (I did because I blindlessly copied the example config in the README), processor.js will return an error which leads to Grover::JavaScript::ReferenceError: type is not defined.

I believe that line 106 it should be `page.emulateVisionDeficiency(visionDeficiency);`

https://github.com/Studiosity/grover/blob/33b05b39086320799feb39bf4ca1daa9e10df353/lib/grover/js/processor.js#L103-L107

I you like, I can make a PR with a fix.

Best regards,
Julian

abrom:
Ahh.. yes that's a typo! I suspect more likely a copy paste out of the documentation.. A PR would be great!

Better yet would be a spec test to validate it's working as expected (i'm a little surprised I hadn't added one myself). The test would seem to be pretty straight forward. If the `vision_deficiency` setting was set to `deuteranopia` then we can just set the background to say red then see that the result of `to_png` returns an image that is an olive type colour

Something like:
```
      context 'when vision deficiency is set to `deuteranopia`' do
        let(:url_or_html) { '<html><body style="background-color: red"></body></html>' }
        let(:options) { { 'visionDeficiency' => 'deuteranopia' } }

        it { expect(convert.unpack('C*')).to start_with "\x89PNG\r\n\x1A\n".unpack('C*') }
        it { expect(image.type).to eq 'PNG' }
        it { expect(image.dimensions).to eq [800, 600] }
        it { expect(mean_colour_statistics(image)).to eq %w[109 90 28] }
      end
```
inserted somewhere around here: https://github.com/Studiosity/grover/blob/main/spec/grover/processor_spec.rb#L827

Please note I've not actually tested those colour statistics.. they are an approximation. But the real numbers should be pretty close to that!

abrom:
Ahh.. actually it looks like the `page.emulateVisionDeficiency` method was only added in v3.2.0

https://github.com/puppeteer/puppeteer/releases/tag/v3.2.0

However I have a helper for that. If you wrap the context in:
```
      # Only test `emulateVisionDeficiency` if the Puppeteer supports it
      if puppeteer_version_on_or_after? '3.2.0'
        context ....
      end
```

abrom:
Fixed in #128

PR #128, opened 2021-08-29
Fix error in processor.js when using the vision_deficiency option

julianwegkamp:
Fixes the error in processor.js when using the vision_deficiency option as reported in issue #127.

Added a spec as you suggested, but I'm running into some strange behavior. On some test runs the mean_colour_statistics returns ["94", "71", "0"] (which I assume is correct) on other runs it returns ["240.506", "240.527", "242.569"]. This is likely the mean color from the page http://www.example.net/foo/bar. Do you recognize this inconsistency while running the specs?

```  
1) Grover::Processor#convert when converting to an image when vision deficiency is set to `deuteranopia` is expected to eq ["94", "71", "0"]
  Failure/Error: it { expect(mean_colour_statistics(image)).to eq %w[94 71 0] }
  
    expected: ["94", "71", "0"]
        got: ["240.506", "240.527", "242.569"]
  
    (compared using ==)
  # ./spec/grover/processor_spec.rb:837:in `block (5 levels) in <top (required)>'
```

### DIFF
--- a/lib/grover/js/processor.js
+++ b/lib/grover/js/processor.js
@@ -103,7 +103,7 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
     // Emulate vision deficiency (if provided)
     const visionDeficiency = options.visionDeficiency; delete options.visionDeficiency;
     if (visionDeficiency !== undefined) {
-      page.emulateVisionDeficiency(type);
+      page.emulateVisionDeficiency(visionDeficiency);
     }
 
     // Bypass CSP (content security policy), if provided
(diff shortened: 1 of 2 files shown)


abrom (approved):
Thanks @julianwegkamp 👍 

FYI I figured out the cause of the weird results you were seeing. It was caused by two different issues. First, it seems like the `emulateVisionDeficiency` method needs to be called *after* the content has been rendered. The documentation doesn't suggest this needs to be the case but I can say that it fails pretty reliably if it isn't afterward! Second, I had missed adding an `await` clause on that call meaning that it was applying the setting asynchronously. This would explain why it would sometimes work and other times not. On the occasions where it worked it was running after the content was rendered. The times when it didn't it appears to have actually broken the rendering mechanism completely. Those numbers coincide with what you see when the renderer doesn't capture the request and it actually goes out to the internet to fetch "example.com". i.e:

foo1

I noticed a few other places where I had not included the `await` clause so I've fixed them up too.

Thanks again

abrom:
FYI released in [v1.0.5](https://rubygems.org/gems/grover/versions/1.0.5)

julianwegkamp:
Thx

PR #131, opened 2021-10-11
Rl feature/support ignore request

braindeaf:
Came across a requirement today to ignore_request? based on the actual request. Sometimes path alone is not enough. This allows testing of the Rack::Request.

```ruby
# config/initializers/grover.rb
Grover.configure do |config|
  # assigning a Proc
  config.ignore_request = ->(req) do
    req.host == 'www.example.com'
  end
  # matches `www.example.com/foo/bar/123.png`

  config.ignore_request = ->(req) do
    req.has_header?('X-BLOCK')
  end
  # matches `HTTP Header X-BLOCK`
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,9 @@ N.B. by default PNG and JPEG are not modified in the middleware to prevent break
 To enable them, there are configuration options for each image type as well as an option to disable the PDF middleware
 (on by default).
 
-If either of the image handling middleware options are enabled, the [ignore_path](#ignore_path) and/or [ignore_request](#ignore_request) should also be configured, otherwise assets are likely to be handled which would likely result in 404 responses.  
+If either of the image handling middleware options are enabled, the [ignore_path](#ignore_path) and/or
+[ignore_request](#ignore_request) should also be configured, otherwise assets are likely to be handled
+which would likely result in 404 responses.  
 
 ```ruby
 # config/initializers/grover.rb
@@ -340,7 +342,7 @@ Grover.configure do |config|
   # matches `www.example.com/foo/bar/123.png`
 
   config.ignore_request = ->(req) do
-    req.headers.has_header?('X-BLOCK')
+    req.has_header?('X-BLOCK')
   end
   # matches `HTTP Header X-BLOCK`
 end
--- a/README.md
+++ b/README.md
@@ -334,12 +334,12 @@ the response. It should be set with a `Proc` which accepts the request (Rack::Re
 # config/initializers/grover.rb
 Grover.configure do |config|
   # assigning a Proc
-  config.ignore_path = ->(req) do
+  config.ignore_request = ->(req) do
     req.host == 'www.example.com'
   end
   # matches `www.example.com/foo/bar/123.png`
 
-  config.ignore_path = ->(req) do
+  config.ignore_request = ->(req) do
     req.headers.has_header?('X-BLOCK')
   end
   # matches `HTTP Header X-BLOCK`
--- a/README.md
+++ b/README.md
@@ -250,12 +250,18 @@ end
 ```
 
 ## Cover pages
+
 Since the header/footer for Puppeteer is configured globally, displaying of front/back cover
 pages (with potentially different headers/footers etc) is not possible.
 
-To get around this, Grover's middleware allows you to specify relative paths for the cover page contents
-via `front_cover_path` and `back_cover_path` either via the global configuration, or via meta tags.
-These paths (with query parameters) are then requested from the downstream app.
+To get around this, Grover's middleware allows you to specify relative paths for the cover page contents.
+For direct execution, you can make multiple calls and combine the resulting PDFs together.
+
+### Using middleware
+
+You can specify relative paths to the cover page contents using the `front_cover_path` and `back_cover_path`
+options either via the global configuration, or via meta tags. These paths (with query parameters) are then
+requested from the downstream app.
 
 The cover pages are converted to PDF in isolation, and then combined together with the original PDF response,
 before being returned back up through the Rack stack.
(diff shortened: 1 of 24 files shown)


spec/grover/middleware_spec.rb
```diff
@@ -860,6 +860,46 @@
...
+            response = get 'http://www.example.org/foobazbar', {}, { 'X-BLOCK' => '1' }
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobazbar.pdf', {}, { 'X-BLOCK' => '1' }
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobarbar', {}, { 'X-BLOCK' => '1' }
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.therealexample.org/foobarbar.pdf'
```
abrom, line 896: ideally this should match the request on L890 (aside from the "block" header) so as to minimise the things changing between success and failure.

spec/grover/middleware_spec.rb
```diff
@@ -860,6 +860,46 @@
...
+          before { allow(Grover.configuration).to receive(:ignore_request).and_return(->(req) { req.has_header?('X-BLOCK') }) }
+
+          it 'request is ignored when the request path passed to the proc defined in ignore_request returns true' do
+            response = get 'http://www.example.org/foobazbar', {}, { 'X-BLOCK' => '1' }
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobazbar.pdf', {}, { 'X-BLOCK' => '1' }
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobarbar', {}, { 'X-BLOCK' => '1' }
```
abrom, line 893: not sure what this is adding to the tests? Ideally this and the next test would both NOT have the extra header. And I would suggest that the URI requested be the same for all (except for the `.pdf` extension where appropriate!)

spec/grover/middleware_spec.rb
```diff
@@ -860,6 +860,46 @@
...
+            response = get 'http://www.example.org/foobazbar'
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobazbar.pdf'
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobarbar'
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.therealexample.org/foobarbar.pdf'
```
abrom, line 878: same applies to these tests. The first two are `www.example.org`, the second two should be something different (not just this one). And I would also suggest that the path (except the extension where appropriate) should be the same. ie `foobazbar[.pdf]`

README.md
```diff
@@ -270,8 +270,7 @@ N.B. by default PNG and JPEG are not modified in the middleware to prevent break
 To enable them, there are configuration options for each image type as well as an option to disable the PDF middleware
 (on by default).
 
-If either of the image handling middleware options are enabled, the [ignore_path](#ignore_path) should also
-be configured, otherwise assets are likely to be handled which would likely result in 404 responses.  
+If either of the image handling middleware options are enabled, the [ignore_path](#ignore_path) and/or [ignore_request](#ignore_request) should also be configured, otherwise assets are likely to be handled which would likely result in 404 responses.
```
abrom, line 273: Can you please split these over multiple lines. Although it doesn't make a difference for the rendered markdown, it's easier to read in most editors if the line length is kept under 120.

```suggestion
If either of the image handling middleware options are enabled, the [ignore_path](#ignore_path) and/or
[ignore_request](#ignore_request) should also be configured, otherwise assets are likely to be handled
which would likely result in 404 responses.  
```

abrom (changes requested):
Change looks good @braindeaf 👍 

A few minor issues with the specs though. Let me know if you need any more info

spec/grover/middleware_spec.rb
braindeaf: OK, I think I got you.

1. blocking with host (html request) returns html
2.blocking with host (pdf request) returns html
3. not blocking with host (html request) returns html
4. not blocking with host (pdf request) returns pdf

spec/grover/middleware_spec.rb
braindeaf: OK, I think I got you.

1. blocking with host (html request) returns html
2.blocking with host (pdf request) returns html
3. not blocking with host (html request) returns html
4. not blocking with host (pdf request) returns pdf

spec/grover/middleware_spec.rb
```diff
@@ -860,6 +860,46 @@
...
+            response = get 'http://www.example.org/foobazbar'
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobazbar.pdf'
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.therealexample.org/foobarbar'
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.therealexample.org/foobarbar.pdf'
```
abrom, line 878:
```suggestion
            response = get 'http://www.therealexample.org/foobazbar'
            expect(response.content_type).to eq 'text/html'

            response = get 'http://www.therealexample.org/foobazbar.pdf'
```

spec/grover/middleware_spec.rb
```diff
@@ -860,6 +860,46 @@
...
+            response = get 'http://www.example.org/foobazbar', {}, { 'X-BLOCK' => '1' }
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.example.org/foobarbar.pdf', {}, { 'X-BLOCK' => '1' }
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.therealexample.org/foobazbar'
+            expect(response.content_type).to eq 'text/html'
+
+            response = get 'http://www.therealexample.org/foobarbar.pdf'
```
abrom, line 896:
```suggestion
            response = get 'http://www.example.org/foobazbar.pdf', {}, { 'X-BLOCK' => '1' }
            expect(response.content_type).to eq 'text/html'

            response = get 'http://www.therealexample.org/foobazbar'
            expect(response.content_type).to eq 'text/html'

            response = get 'http://www.therealexample.org/foobazbar.pdf'
```

braindeaf:
Excellent, thank you :)

PR #147, opened 2021-12-18
Ruby 3.1 compatibility

walski:
Ruby 3.1's release is around the corner and grover seems to work just fine with the Ruby 3.1.0-preview1. Specs all pass.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,7 @@ require: rubocop-rspec
 
 AllCops:
   NewCops: enable
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 2.6
 
 Layout/DotPosition:
   EnforcedStyle: trailing
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,15 +17,15 @@ env:
     - PUPPETEER_VERSION=11.0.0
     - PUPPETEER_VERSION=12.0.1
 
-rvm: 3.0
+rvm: 3.1.0
 
 jobs:
   include:
-    - rvm: 2.5
-      env: PUPPETEER_VERSION=12.0.1
     - rvm: 2.6
       env: PUPPETEER_VERSION=12.0.1
     - rvm: 2.7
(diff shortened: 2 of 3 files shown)


abrom:
Thanks for the PR @walski 👍 

Is there any urgency in applying/releasing this? I'd prefer to have the new Ruby version included in the CI build process before release but that would require the new Ruby version to be released via RVM. Another option would be to include `ruby-head` in the list of Rubies.. although this would potentially add an air of flake to the CI..

walski:
No urgency at all. Just needed it to prepare our app for 3.1 and thought I'd open the PR if the fork is there, already :) Have a great Christmas season and thanks for your work!

abrom:
Looks like although RVM does support Ruby 3.1, they haven't included pre-built binaries for it yet. I don't really want to skip using binaries as building from source in CI will take an age.. I'll keep an eye on it and rebuild once the binaries are available 😄

entretechno-jeremiah:
> Looks like although RVM does support Ruby 3.1, they haven't included pre-built binaries for it yet. I don't really want to skip using binaries as building from source in CI will take an age.. I'll keep an eye on it and rebuild once the binaries are available :smile:

Don't hold your breath. They haven't published any binaries in a little over a year now haha. (They have Ruby 3.0.0 but not 3.0.1-3.0.3.)

akaspick:
@entretechno-jeremiah Yeah, I don't see a 3.1 binary getting released anytime soon via rvm. I think this needs to move ahead without that as plan A and time to look at a plan B.

walski:
probably about time to move off Travis

abrom:
I've updated it to use `ruby-head` which was last built in the Travis RVM mirror at the end of the year (so is going to be pretty close to the 3.1 release).

Yes, I agree.. moving away from Travis is on my TODO list

abrom:
Released as [v1.1.1](https://rubygems.org/gems/grover/versions/1.1.1)

walski:
🙏  thank you so much!

PR #160, opened 2022-09-05
fix: support projects that are using esm instead of cjs

ElMassimo:
### Description 📖 

This pull request fixes an error that occurs when using Grover in ESM applications:

```
Grover::Error
ReferenceError: require is not defined
```

### Background 📜

Node.js has two module systems, [CommonJS](https://nodejs.org/api/modules.html) modules and [ECMAScript modules](https://nodejs.org/api/esm.html#modules-ecmascript-modules).

The [`processor.js`](https://github.com/Studiosity/grover/blob/main/lib/grover/js/processor.js) script targets the `CJS` dialect, and uses `require` to import libraries.

Since Grover executes the script in the context of the parent app directory, Node.js will infer the default module system from the app's `package.json`.

When using Grover in apps with `"type": "module"`, Node.js loads `processor.js` in ESM mode, failing with:

```
ReferenceError: require is not defined
```

### The Fix 🔨

Ensure the script is always executed by Node.js using the CJS module system.

By renaming `processor.js` to `processor.cjs`, Node.js will always interpret the file as CJS, regardless of the configuration of the app's `package.json`.

This fix preserves compatibility with CJS apps, so it's not a breaking change.


abrom (approved):
Looks great, thanks @ElMassimo 👍

abrom:
Released in [v1.1.2](https://rubygems.org/gems/grover/versions/1.1.2)

PR #165, opened 2022-10-17
:wrench: Support pipe: true setting

gretch:


abrom:
Any movement on this @gretch  ?

Issue #166, opened 2022-11-13
Unable to run it on Heroku


pvaldesoiro:
Hello Grover team,

First of all, thank you for your work on this tool.

I have been trying to run Grover in a rails app to generate PDFs from HTML (not from views, but from standalone HTMLs). When I run it locally, it runs perfectly. But when I run it on the server (deployed to Heroku), it's impossible for me to make it work. It gives me a runtime error:
```
2022-11-13T22:43:47.411370+00:00 app[web.1]: Grover::JavaScript::Error (Could not find Chromium (rev. 1056772). This can occur if either

2022-11-13T22:43:47.411371+00:00 app[web.1]:  1. you did not perform an installation before running the script (e.g. `npm install`) or

2022-11-13T22:43:47.411374+00:00 app[web.1]:  2. your cache path is incorrectly configured (which is: /app/.cache/puppeteer).

2022-11-13T22:43:47.411374+00:00 app[web.1]: For (2), check out our guide on configuring puppeteer at https://pptr.dev/guides/configuration.):
```

Project stack:
- Grover 1.1.2
- Ruby 3.1.2
- Rails 6.1.7
- Node 18.12.1
- Puppeteer 19.2.2 (in package.json)

Buildpacks in order:
- heroku/nodejs
- https://buildpack-registry.s3.amazonaws.com/buildpacks/jontewks/puppeteer.tgz
- heroku/ruby

Environment variable:
- GROVER_NO_SANDBOX=true

My actions:
- I have tried to purge my Heroku cache as recommended in the [Puppeteer buildpack](https://elements.heroku.com/buildpacks/jontewks/puppeteer-heroku-buildpack).
- I have followed the requirement that was shown in a previous error of adding the Puppeteer dependency to my package.json (through yarn).
- I have purged cache again.
- I have tried several Node versions (16 and 18)
- I don't have any puppeteer configuration file, any javascript code, nor have I configured any cache path.

Thank you for your help.

pvaldesoiro:
Okay, I found the problem. There was a [breaking change](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/CHANGELOG.md#1900-2022-10-14) in Puppeteer 19.0.0.

**I fixed it by downgrading to Puppeteer 18.2.1.**

[...]

PR #167, opened 2022-11-14
Extend "Running on Heroku" readme section.

pvaldesoiro:
Added a step to add Puppeteer to package.json, and a warning about incompatibility with Heroku and certain versions of Puppeteer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,18 +422,24 @@ require 'combine_pdf'
 ## Running on Heroku
 
 To run Grover (Puppeteer) on Heroku follow these steps:
-1. Add the `node` buildpack. Puppeteer requires a node environment to run.
+1. Add Puppeteer to package.json.
+   ```
+   npm install puppeteer@18.2.1
+   ```
+   > <sub>⚠ It is highly recommended to use Puppeteer '<=18.2.1', since '19.0.0 >= 19.2.2' can give errors in Heroku.</sub>
+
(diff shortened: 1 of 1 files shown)


abrom:
Hi @pvaldesoiro 

Thank you for the PR, although per my comment in the issue you raised about this https://github.com/Studiosity/grover/issues/166#issuecomment-1313378539 I suspect there is likely a better way to keep things moving forward rather than requiring puppeteer be pegged at a specific version. I'd appreciate your thoughts and any feedback you can give me on the above tagged comment!

pvaldesoiro:
Hello @abrom, sorry for the late reply.

I think you are right. Adding a warning informing the user about adding `.puppeteerrc.cjs` with the cache config in case they want to make Grover work with puppeteer v19.x would be a much better option than advising to keep puppeteer stuck at an old version. I will update the PR if you are fine with it. But anyway, if you have any other suggestions I would have no problem with trying an alternative approach; or you prefer to keep the readme as it is right now and address the issue in future versions, I can just close the PR.

abrom:
Sounds great, updating the README to include details about the config change feels like the best option. Thanks @pvaldesoiro 👍

abrom:
Good to close this now that #169 is merged?

pvaldesoiro:
@abrom Yes. Closed.

PR #169, opened 2022-12-22
Include instructions for .cjs file for heroku

MatthewSchultz:
Include suggestion from @abrom to enable this on Heroku using 19+.


abrom:
Relates to #166

abrom (approved):
Nice one, thanks @MatthewSchultz

PR #170, opened 2022-12-28
Ruby `3.2.0` compatibility

Yobilat:
**Changes:**

- Added compatibility for ruby 3.2.0
- Tested with newer versions of puppeteer
- Miscellaneous (rubocop, manual) fixes
- Github actions update

**Questions:**

- Do we need to test puppeteer as far as v1 given still suported nodejs versions?
  - If not can we clear up tests that depends on puppeteer version (`puppeteer_version_on_or_after?`)? 
- Some [tests](https://github.com/Studiosity/grover/pull/170/files#diff-beedb588836752a1b300209c575d884951d3cca658813e67c9d1dad7164babf1L367) depends on [external resources](https://cookie-renderer.herokuapp.com/) which are now failing - probably because heroxu axed free dynos. Any sugestions if/how to replace that?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,6 +52,6 @@ jobs:
           PUPPETEER_VERSION: ${{ matrix.puppeteer-version }}
 
       - name: Test & publish code coverage
-        uses: paambaati/codeclimate-action@v3
+        uses: paambaati/codeclimate-action@v3.2.0
         env:
           CC_TEST_REPORTER_ID: 5cfed40102c670b5c9e509730782b751939ddbe53fc57c317b718f635bab1ce8
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,52 +12,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: ['3.1']
-        node-version: ['10']
-        puppeteer-version: [
-            '1.20.0',
-            '2.1.1',
-            '3.3.0',
-            '4.0.1',
-            '5.5.0',
-            '6.0.0',
-            '7.1.0',
-            '8.0.0',
-            '9.1.1',
-            '10.4.0',
-            '11.0.0',
-            '12.0.1',
-            '13.7.0',
-            '14.4.1',
-            '15.5.0',
-            '16.2.0',
-            '17.1.3'
-        ]
+        ruby-version: ['3.2']
+        node-version: ['14', '16', '18']
+        puppeteer-version: ['14.4.1', '15.5.0', '16.2.0', '17.1.3', '18.2.1', '19.4.1']
         include:
-          - ruby-version: '2.6'
-            node-version: '10'
-            puppeteer-version: '17.1.3'
-          - ruby-version: '2.7'
-            node-version: '10'
-            puppeteer-version: '17.1.3'
-          - ruby-version: '3.0'
-            node-version: '10'
-            puppeteer-version: '17.1.3'
-          - ruby-version: '3.1'
-            node-version: '12'
-            puppeteer-version: '17.1.3'
           - ruby-version: '3.1'
             node-version: '14'
-            puppeteer-version: '17.1.3'
-          - ruby-version: '3.1'
-            node-version: '16'
-            puppeteer-version: '17.1.3'
-          - ruby-version: '3.1'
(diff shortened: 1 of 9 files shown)


abrom:
> Do we need to test puppeteer as far as v1 given still suported nodejs versions?
> If not can we clear up tests that depends on puppeteer version (puppeteer_version_on_or_after?)?

No, likely safe to start dropping some of the older ones. Once an older version is dropped then it would be safe to remove the guard around any of the `puppeteer_version_on_or_after` tests (but the tests themselves shouldn't be removed!!?)

> Some [tests](https://github.com/Studiosity/grover/pull/170/files#diff-beedb588836752a1b300209c575d884951d3cca658813e67c9d1dad7164babf1L367) depends on [external resources](https://cookie-renderer.herokuapp.com/) which are now failing - probably because heroxu axed free dynos. Any sugestions if/how to replace that?

Ahh.. yes. I'll need to fix that, thanks.

abrom:
yes, although there is another issue with a change to how newer versions of rspec deal with hash vs keyword arguments I still need to fix up 😬

cpjmcquillan:
The rspec-mocks hash vs keyword arguments issue with Ruby 3.2 is resolved in 3.12.2, so you shouldn't have to change anything in your tests so long as rspec has been updated

cpjmcquillan:
> I've already fixed tests regarding that.

Sorry, I didn't mean to chime in and detract from the work happening here.

I only wanted to point out the issue described in https://github.com/rspec/rspec-mocks/issues/1513 was resolved in https://github.com/rspec/rspec-mocks/pull/1514, and as far as I know tests shouldn't need to be fixed as long as `rspec-mocks` has been updated to the latest version.

---

On another note, would it be possible to relax the `required_ruby_version` constraint? 

I'm currently blocked from updating Ruby version by this requirement. I understand not wanting to support older versions of Ruby and only wanting to explicitly support certain versions at a given time, but this feels overly restrictive. 

It also adds some pressure to maintainers when new versions are released that could be avoided.

How do you feel about adjusting the requirement to something like this instead?

```ruby
spec.required_ruby_version = ">= 2.7.0"
```

abrom:
> > I've already fixed tests regarding that.
> 
> Sorry, I didn't mean to chime in and detract from the work happening here.
> 
> I only wanted to point out the issue described in [rspec/rspec-mocks#1513](https://github.com/rspec/rspec-mocks/issues/1513) was resolved in [rspec/rspec-mocks#1514](https://github.com/rspec/rspec-mocks/pull/1514), and as far as I know tests shouldn't need to be fixed as long as `rspec-mocks` has been updated to the latest version.

That's not the case. The tests run against the latest versions of `rspec` and `rspec-mocks`. See https://github.com/Studiosity/grover/actions/runs/3870089736/jobs/6596725310


> 
> On another note, would it be possible to relax the `required_ruby_version` constraint?
> 
> I'm currently blocked from updating Ruby version by this requirement. I understand not wanting to support older versions of Ruby and only wanting to explicitly support certain versions at a given time, but this feels overly restrictive.
> 
> It also adds some pressure to maintainers when new versions are released that could be avoided.
> 
> How do you feel about adjusting the requirement to something like this instead?
> 
> ```ruby
> spec.required_ruby_version = ">= 2.7.0"
> ```
> 
> Thanks to everyone that works on this gem, your work is appreciated ❤️

Except then you have the situation where someone upgrades to the latest version.. gem installs fine, thus assumes everything must be fine. Except what if it isn't. If you have unbounded version checks you leave yourself very open to the unexpected/breakages/bugs/crashes/etc. It's why the project has the version matrix testing the code in CI.

Short term option.. fork the project, update the required version to your liking, use the fork in your project until grover + 3.2.x has been properly tested. 😄 But of course that goes with the caveat that it hasn't been properly tested.

breisig:
We REALLY need Ruby 3.2 support.

abrom:
FYI @Yobilat I've released Ruby 3.2.x grover support in [v1.1.3](https://rubygems.org/gems/grover/versions/1.1.3)

Thank you for your efforts in getting this through, although note I stripped parts of your PR out into separate PRs. It was trying to do a little too much all in one go, and I wasn't a fan of your matrix re-org.. it would have meant a full 3 (node versions) X 6 (puppeteer versions) + 3 static builds per run.

PR #178, opened 2023-01-24
Implement remote browser option

Zinggi:
Idea taken from: https://github.com/Studiosity/grover/pull/115 from @drnic

---

Can be used with

```ruby
Grover.new(..., browser_ws_endpoint: "ws://localhost:3333")
```

and e.g. a `docker-compose.yaml` file like:
```yaml
version: "3.7"
services:
  chrome:
    image: browserless/chrome:1.57-puppeteer-19.2.2
    ports:
      - 3333:3000
```

Hopefully this PR is much easier to review and merge than the original one where I took the idea from.

We have been using this in a production system for a couple of months and haven't had any issues with this setup.

---

I can include an update to the docs with this PR as well, but I wasn't sure where exactly I should put this.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,11 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    services:
+      remote-chrome:
+        image: browserless/chrome:1.57-puppeteer-19.2.2
+        ports:
+          - 3000:3000
     strategy:
       fail-fast: false
       matrix:
@@ -61,6 +66,11 @@ jobs:
 
       - name: Run tests
         run: bundle exec rspec
+        env:
+          PUPPETEER_VERSION: ${{ matrix.puppeteer-version }}
+
+      - name: Run tests with remote browser
+        run: bundle exec rspec --tag remote-browser
         env:
           PUPPETEER_VERSION: ${{ matrix.puppeteer-version }}
 
--- a/README.md
+++ b/README.md
@@ -179,6 +179,21 @@ only really makes sense if you're calling Grover directly (and not via middlewar
 Grover.new('<some URI with basic authentication', username: 'the username', password: 'super secret').to_pdf
 ```
 
+#### Remote Chromium
+
+By default, Grover launches a local Chromium instance. You can connect to a remote/external
+Chromium with the `browser_ws_endpoint` options.
+
+For example, to connect to a chrome instance started with docker using `docker run -p 3000:3000 browserless/chrome:latest`:
+
+```ruby
+options = {"browser_ws_endpoint": "ws://localhost:3000/"}
+grover = Grover.new("https://mysite.com/path/to/thing", options)
+File.open("grover.png", "wb") { |f| f << grover.to_png }
+```
+
+You can also pass launch flags like this: `ws://localhost:3000/?--disable-speech-api`
+
 #### Adding cookies
 To set request cookies when requesting a URL, pass an array of hashes as such
 _N.B._ Only the `name` and `value` properties are required.
--- a/README.md
+++ b/README.md
@@ -490,6 +490,12 @@ $ rubocop
 both succeed
 
 
+To run tests tagged with `remote_browser`, you need to start a browser in a container:
+`docker run -p 3000:3000 browserless/chrome:latest`
+and run:
+`rspec --tag remote_browser`
+
+
 ## Special mention
 Thanks are given to the great work done in the [PDFKit project](https://github.com/pdfkit/pdfkit).
 The middleware and HTML preprocessing components were used heavily in the implementation of Grover.
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -17,13 +17,13 @@ const path = require('path');
 const _processPage = (async (convertAction, urlOrHtml, options) => {
   let browser, errors = [], tmpDir;
 
-  // Configure puppeteer debugging options
-  const debug = options.debug; delete options.debug;
-  const remoteConnect = typeof options.browserWsEndpoint === "string";
   try {
-    if (remoteConnect) {
+    // Configure puppeteer debugging options
+    const debug = options.debug; delete options.debug;
+    const browserWsEndpoint = options.browserWsEndpoint; delete options.browserWsEndpoint;
+    if (typeof browserWsEndpoint === "string") {
       const connectParams = {
-        browserWSEndpoint: options.browserWsEndpoint,
+        browserWSEndpoint: browserWsEndpoint,
       };
 
       browser = await puppeteer.connect(connectParams);
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -16,35 +16,46 @@ const path = require('path');
 
 const _processPage = (async (convertAction, urlOrHtml, options) => {
   let browser, errors = [], tmpDir;
+
+  // Configure puppeteer debugging options
+  const debug = options.debug; delete options.debug;
+  const remoteConnect = typeof options.browserWsEndpoint === "string";
   try {
-    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'grover-'));
+    if (remoteConnect) {
+      const connectParams = {
+        browserWSEndpoint: options.browserWsEndpoint,
+      };
 
-    const launchParams = {
-      args: process.env.GROVER_NO_SANDBOX === 'true' ? ['--no-sandbox', '--disable-setuid-sandbox'] : [],
-      userDataDir: tmpDir
-    };
+      browser = await puppeteer.connect(connectParams);
+    } else {
+      tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'grover-'));
 
-    // Configure puppeteer debugging options
-    const debug = options.debug; delete options.debug;
-    if (typeof debug === 'object' && !!debug) {
-      if (debug.headless !== undefined) { launchParams.headless = debug.headless; }
-      if (debug.devtools !== undefined) { launchParams.devtools = debug.devtools; }
+      const launchParams = {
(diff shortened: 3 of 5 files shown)


visini:
Great work, thank you @Zinggi

@abrom What are your thoughts on this?

Zinggi:
I just rebased to solve conflicts, thanks for the reminder @visini :)

Hint for the review: The diff looks a lot cleaner if you enable "hide whitespace" setting:

Screenshot from 2023-02-23 11-42-01

Zinggi:
Changed to draft, as I made an error while rebasing. Will fix and mark ready again when correct

Zinggi:
All correct again, sorry for the bad rebase from before.
The current version has now also been tested on our production system where we still use my fork.

lib/grover/js/processor.cjs
```diff
@@ -16,35 +16,46 @@ const path = require('path');
...
-      if (debug.devtools !== undefined) { launchParams.devtools = debug.devtools; }
-    }
 
-    // Configure additional launch arguments
-    const args = options.launchArgs; delete options.launchArgs;
-    if (Array.isArray(args)) {
-      launchParams.args = launchParams.args.concat(args);
-    }
+  // Configure puppeteer debugging options
+  const debug = options.debug; delete options.debug;
```
abrom, line 21: seems like these two have been moved away from the rest of the related code? Not clear why they shouldn't be down at ~L38?

lib/grover/js/processor.cjs
```diff
@@ -16,35 +16,46 @@ const path = require('path');
...
+  const remoteConnect = typeof options.browserWsEndpoint === "string";
```
abrom, line 22: feels like this should live inside the `try` block

lib/grover/js/processor.cjs
```diff
@@ -16,35 +16,46 @@ const path = require('path');
...
+  try {
+    if (remoteConnect) {
+      const connectParams = {
+        browserWSEndpoint: options.browserWsEndpoint,
+      };
```
abrom, line 27: i'd recommend the pattern:
```suggestion
    const browserWsEndpoint = options.browserWsEndpoint; delete options.browserWsEndpoint;
    if (typeof browserWsEndpoint === "string") {
      const connectParams = {
        browserWSEndpoint: browserWsEndpoint,
      };
```
The main reason is because the options are simply passed through verbatim to the convert action (`pdf`, `content` or `screenshot`). Safer to remove these sorts of custom options as we check/use them to prevent any sort of unexpected behaviour elsewhere!

abrom (changes requested):
As with the PR from @drnic, I'm all for the change but I'd really like to see some form of automated test that could go with it. Of course one of the main issues there would be how we can configure the tests to use some browser endpoint.. As outlined in #115 it might be possible to use browserless.io for that purpose? Thoughts? 

It would also be good to see some documentation to explain to others how to use the feature. #115 included something that would likely be ideally suited.

lib/grover/js/processor.cjs
Zinggi: It's used again at the very end of the `try { }` block. Moving it inside the else makes it undefined when using remote browser option

Zinggi:
Hi @abrom, thank you for the review!
I updated my PR to include your feedback.

As for testing, my suggestion is to just start a chrome container on the CI and use that one.
Not sure about a testing matrix there, maybe just one is enough?
I also suggest to not use ENV values to check if we are running in CI, but instead make use of rspec tags to exclude those tests by default.
My latest commit tries to do that.

lib/grover/js/processor.cjs
abrom: Of course, my mistake 😄

Zinggi:
Just saw the rubocop issue, it's now fixed

abrom:
FYI the tests are currently failing because my cookie/header test system is offline 😞 I'll have it back online tomorrow

spec/grover/processor_spec.rb
```diff
@@ -875,6 +875,16 @@
...
+      context 'when using remote browser option with HTML', remote_browser: true do
+        let(:options) { { browserWsEndpoint: 'ws://localhost:3000' } }
+        let(:url_or_html) { '<html><body style="background-color: blue"></body></html>' }
+
+        it { expect(convert.unpack('C*')).to start_with "\x89PNG\r\n\x1A\n".unpack('C*') }
+        it { expect(image.type).to eq 'PNG' }
+        it { expect(image.dimensions).to eq [800, 600] }
+        it { expect(mean_colour_statistics(image)).to eq %w[0 0 255] }
+      end
+
```
abrom, line 887: I'm not convinced this is really proving that the WS endpoint stuff is actually working. If you were to revert the changes in the processor.cjs file these tests would still pass. Suggest moving remove this image based test, and instead use something like this up in the PDF test section:
```
        context 'when passing through launch params' do
          let(:options) { { 'browserWsEndpoint' => browser_ws_endpoint } }
          let(:browser_ws_endpoint) { 'ws://localhost:3000/' }
          let(:url_or_html) do
            <<-HTML
              <html>
                <body>
                  Speech recognition is <span id="test" />
                  <script type="text/javascript">
                    var speechSupported = "webkitSpeechRecognition" in window;
                    document.getElementById("test").innerHTML = speechSupported ? "supported" : "not supported"
                  </script>
                </body>
              </html>
            HTML
          end

          it { expect(pdf_text_content).to eq 'Speech recognition is supported' }

          context 'when WS endpoint param specifies disabling the speech API' do
            let(:browser_ws_endpoint) { 'ws://localhost:3000/?--disable-speech-api' }

            it { expect(pdf_text_content).to eq 'Speech recognition is not supported' }
          end
        end
```
That proves that it is the browser WS endpoint configuration that is changing the text content. Win!

README.md
```diff
@@ -490,6 +490,12 @@ $ rubocop
 both succeed
 
 
+To run tests tagged with `remote_browser`, you need to start a browser in a container:
+`docker run -p 3000:3000 browserless/chrome:latest`
+and run:
+`rspec --tag remote_browser`
```
abrom, line 496: definitely good to have this in the README, but it would also be great to have some explanation on how to use this new feature too. You can likely just pull the update from #115 ie https://github.com/Studiosity/grover/pull/115/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R165-R176

spec/grover/processor_spec.rb
Zinggi: Hmm, I know they worked because I saw the logs from the docker container.
But I get your point, reverting the changes would still make this pass.

I'll try your suggestion when I next get to it, thanks.

Zinggi: I added the suggested test as a new commit.
Indeed this one does fail when undoing the changes, while the others still pass.

Should I remove the other tests or should I keep them anyway?

.github/workflows/test.yml
```diff
@@ -64,6 +69,11 @@ jobs:
         env:
           PUPPETEER_VERSION: ${{ matrix.puppeteer-version }}
 
+      - name: Run tests with remote browser
```
Zinggi, line 72: this could be run in parallel to the other test, but it might not be worth it. But maybe something to try if this becomes too slow.

.github/workflows/test.yml
```diff
@@ -9,6 +9,11 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    services:
+      remote-chrome:
+        image: browserless/chrome:1.57-puppeteer-19.2.2
```
Zinggi, line 14: maybe it would make sense to also use a version matrix here, but I have no experience with that..

abrom (approved):
Nice one @Zinggi looks great 👍

Zinggi:
@abrom Thanks for the thorough review, I learned something about proper testing thanks to your testing feedback :+1:

abrom:
No worries, yes testing is always about the edge cases.. how does it behave when something is "off" vs "on" (or similar). Same goes for, will this new test pass if I change the code back. Always need to assume that the API of the thing you're using is going to break in some way you aren't expecting 😆

abrom:
Doh.. looks like there was a typo in the "include" option: 
```
Run options:
  include {:"remote-browser"=>true}
  exclude {:remote_browser=>true}

Randomized with seed 6018[7](https://github.com/Studiosity/grover/actions/runs/4314194794/jobs/7526914798#step:9:8)


Finished in 0.00034 seconds (files took 0.5963[8](https://github.com/Studiosity/grover/actions/runs/4314194794/jobs/7526914798#step:9:9) seconds to load)
0 examples, 0 failures
```

abrom:
Note, I've got the fix for it building now.. If that all passes I'll get a gem cut

abrom:
That's better! 
```
Run bundle exec rspec --tag remote_browser
  
Run options: include {:remote_browser=>true}

Randomized with seed 24253
......

Finished in 8.74 seconds (files took 0.42484 seconds to load)
[6](https://github.com/Studiosity/grover/actions/runs/4314415511/jobs/7527409122#step:9:6) examples, 0 failures
```

krzysztofjablonski:
@Zinggi, I've tried to use the browser_ws_endpoint with a running Chrome instance, but every time I utilize Grover, Chrome gets closed. This probably happens because of the following code:
```
    if (browser) {
      await browser.close();
    }
```

Do you have any tips on how to make it work without Chrome closing each time?

Zinggi:
> ...
> Do you have any tips on how to make it work without Chrome closing each time?

Sorry, I have no idea.
This never happened for me. But I'm using chromium, maybe that makes a difference?
But it seems unlikely that `browser.close()` would close your browser, if using `browser_ws_endpoint`, as then `browser` should refer to the puppeteer controlled instance (in my case `browserless/chrome`).

krzysztofjablonski:
@Zinggi it seems this PR solves my problems: https://github.com/Studiosity/grover/pull/183

PR #179, opened 2023-02-05
Clean up temporary user data directory

abrom:
If not specified, the temporary user data directory is not cleaned up automatically. This can eventually result in the filesystem getting filled with puppeteer chrome profile files.


micred:
I got a cluttered temporary directory even with grover 1.1.5 and I noticed that in that system there was an old node 12 and it was throwing the following error:
```
fs.rmSync is not a function
```
I will update node.

PR #183, opened 2023-03-04
Puppeteer core and remote browser

dexterouschen:
- keep remote browser running since it may be shared by others
- support for puppeteer-core feels natural if remote browser is supported

### DIFF
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -2,7 +2,12 @@
 try {
   const Module = require('module');
   // resolve puppeteer from the CWD instead of where this script is located
-  var puppeteer = require(require.resolve('puppeteer', { paths: Module._nodeModulePaths(process.cwd()) }));
+  var puppeteer;
+  try {
+    puppeteer = require(require.resolve('puppeteer', { paths: Module._nodeModulePaths(process.cwd()) }));
+  } catch (e) {
+    puppeteer = require(require.resolve('puppeteer-core', { paths: Module._nodeModulePaths(process.cwd()) }));
+  }
 } catch (e) {
   process.stdout.write(JSON.stringify(['err', e.toString()]));
   process.stdout.write("\n");
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -15,12 +15,12 @@ const os = require('os');
 const path = require('path');
 
 const _processPage = (async (convertAction, urlOrHtml, options) => {
-  let browser, errors = [], tmpDir;
+  let browser, page, browserWsEndpoint, errors = [], tmpDir;
 
   try {
     // Configure puppeteer debugging options
     const debug = options.debug; delete options.debug;
-    const browserWsEndpoint = options.browserWsEndpoint; delete options.browserWsEndpoint;
+    browserWsEndpoint = options.browserWsEndpoint; delete options.browserWsEndpoint;
     if (typeof browserWsEndpoint === "string") {
       const connectParams = {
         browserWSEndpoint: browserWsEndpoint,
@@ -56,7 +56,7 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
       browser = await puppeteer.launch(launchParams);
     }
 
-    const page = await browser.newPage();
+    page = await browser.newPage();
 
     // Basic auth
     const username = options.username; delete options.username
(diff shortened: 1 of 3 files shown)


lib/grover/js/processor.cjs
```diff
@@ -2,7 +2,12 @@
 try {
   const Module = require('module');
   // resolve puppeteer from the CWD instead of where this script is located
-  var puppeteer = require(require.resolve('puppeteer', { paths: Module._nodeModulePaths(process.cwd()) }));
+  var puppeteer;
+  try {
+    puppeteer = require(require.resolve('puppeteer', { paths: Module._nodeModulePaths(process.cwd()) }));
+  } catch (e) {
+    puppeteer = require(require.resolve('puppeteer-core', { paths: Module._nodeModulePaths(process.cwd()) }));
```
abrom, line 9: i'm not convinced this solves the problem without creating a new one. With your change the warning people will see if/when puppeteer isn't installed is that they need to install `puppeteer-core`, which is *not* the desired behaviour. Have you run the specs to test this? I have two specs testing exactly that and I don't see them passing with your change.

While on specs, this functionality is testable and I'd like to see any change include tests to show it works the way expected to prevent regressions.

abrom (changes requested):
In theory the change is good, but it seems to have created a new issue, and there are no tests .

lib/grover/js/processor.cjs
dexterouschen: @abrom since `options.browserWsEndpoint` is needed to determine the proper dependency, can we move the `require`s either in `_processPage` or further down in the listener function?

dexterouschen: @abrom I made a new commit that passed all tests including the remote ones, I did have to change a lot more code though. let me know what you think.

lib/grover/js/processor.cjs
```diff
@@ -287,7 +284,30 @@ require('readline').createInterface({
...
+    // Setup imports
+    try {
+      const Module = require('module');
+
+      // resolve puppeteer from the CWD instead of where this script is located
+      try {
+        puppeteer = require(require.resolve('puppeteer', { paths: Module._nodeModulePaths(process.cwd()) }));
+      } catch (e) {
+        if (typeof options.browserWsEndpoint === "string") {
+          puppeteer = require(require.resolve('puppeteer-core', { paths: Module._nodeModulePaths(process.cwd()) }));
```
abrom, line 299: hi @dexterouschen I see you've moved the import into the page processor, but that still doesn't fix the root problem.. and it creates a new problem of the JS script returning to the RB that the code loaded ok when it hasn't loaded yet?! Ie see line 1 of this file. It is returning `ok` when it definitely shouldn't. 

I would see the solution as simply capturing the original load failure, then throwing that when the inner load failure occurs. ie
```
try {
  const Module = require('module');

  try {
    // resolve puppeteer from the CWD instead of where this script is located
    var puppeteer = require(require.resolve('puppeteer', { paths: Module._nodeModulePaths(process.cwd()) }));
  } catch (puppeteerError) {
    try {
      // try resolve `puppeteer-core` library instead
      var puppeteer = require(require.resolve('puppeteer-core', { paths: Module._nodeModulePaths(process.cwd()) }));
    } catch (coreError) {
      // raise the original puppeteer load issue so we don't send people don't the wrong rabbit hole by default.
      throw puppeteerError;
    }
  }
} catch (e) {
  process.stdout.write(JSON.stringify(['err', e.toString()]));
  process.stdout.write("\n");
  process.exit(1);
}

Comment (sbounmy) [reply lib/grover/js/processor.cjs:299]: @abrom was this part not merged? 
I have
```js
Cannot find module 'puppeteer'. You need to add it to '/app/package.json' and run 'npm install'
```
error because I am using
`puppeteer-core` with a remote docker `alpine-chrome:119` via `browser_ws_endpoint`

```
rover.configure do |config|
  config.options = {
    # executable_path: ENV['PUPPETEER_EXECUTABLE_PATH'],
    launch_args: ['--disable-dev-shm-usage', '--no-sandbox', '--disable-dev-shm-usage'],
    browser_ws_endpoint: ' ws://0.0.0.0:9222/devtools/browser/xxx',
  }
end
```

abrom: No. The author closed the PR 1 month ago.

sbounmy: @abrom just did a PR for that :) https://github.com/Studiosity/grover/pull/225

PR #186, opened 2023-03-27
Fix Rubocop lint

petergoldstein:
Fixed the newly failing lint that was blocking running tests.

Note that tests are still failing.  It looks like an issue with `https://cookierenderer-production.up.railway.app/`


abrom (approved):
Thanks @petergoldstein 👍

Issue #162, opened 2022-09-23
Would you consider to migrate from puppeter/node to ferrum/ruby


pftg:
There is node dependency here, but this dependency could be avoided by using https://github.com/rubycdp/ferrum.

This will flex dependencies, and simplify installation and testing.

abrom:
Sorry for the late reply @pftg, I must have missed the notification.

Note I have researched this in the past but found the Ruby CDP options out there were significantly slower than using Puppeteer (I tried a couple). It was a while ago, but I recall it being in the order of 30-40% slower.

Also, whilst I am by no means suggesting that those packages aren't well maintained, the Puppeteer library is maintained by Google so is likely to contain all the latest features and be well supported in the case of bugs etc.

But if you'd prefer use Ferrum or similar you are of course welcome to fork the project! 😄

pftg:
Yeah, the support fork would be a little harder

NARKOZ:
For a PDF generator using Ferrum check https://github.com/excid3/ferrum_pdf

PR #187, opened 2023-03-27
[WIP] Add Ferrum backend?

botandrose:
Hello!

This PR swaps out Puppeteer for https://github.com/rubycdp/ferrum in an extremely basic way. I don't think it should be merged as-is, of course, but I wanted to see if there's an appetite for adding Ferrum as a second backend, perhaps. If so, I'd be willing to work this into a proper PR.

The main impetus behind this was the desire to remove the Node.js dependency from our Rails apps, and Grover was our final usage of it. I love Grover, but not so much Node.js, so I tried replacing Puppeteer with Ferrum, which is in pure Ruby, and it was shockingly easy! At least for our specific use-case.

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ Gemfile.lock
 package-lock.json
 /node_modules
 /coverage
+/.ruby-version
+/.nvmrc
+/.byebug_history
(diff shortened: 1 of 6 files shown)


abrom:
Hi @botandrose 

I've had this come up before, and I've investigated libraries (including Ferrum) myself in the past. In short, it/they don't work as well as Puppeteer.

See also #162

HTH

botandrose:
Roger that, @abrom. Thanks for the reply! I'll go ahead and close this PR.

abrom:
I just looked at the end to end performance/timing of content load + conversion. Digging into why wasn't really something I was interested in

PR #225, opened 2024-02-12
Puppeteer core dependency

sbounmy:
when using remote chromium `browserless/chrome`, `puppeteer-core` is more lightweight and avoid download chrome.


rud (approved):
Sounds like an excellent plan! 

An idea: Maybe make it a config option so one can explicitly pick whether `puppeteer` or `puppeteer-core` is the command to invoke?

Note: I'm not affiliated with the project, but this seems like a good plan.

sbounmy:
@rud IMHO its better to not have another config option.
Most people have puppeteer or puppeteer-core installed so the gem should be ale to pick which one is available.
`puppeteer` or fallback to `puppeteer-core`

sbounmy:
@abrom any update for merging this ?

spec/grover/processor_spec.rb
```diff
@@ -95,6 +101,23 @@
...
+          FileUtils.move 'node_modules/puppeteer_temp', 'node_modules/puppeteer'
+        end
+
+        let(:options) { { 'browserWsEndpoint' => browser_ws_endpoint } }
+        let(:browser_ws_endpoint) { 'ws://localhost:3000/' }
+        let(:url_or_html) { '<html><body style="background-color: blue"></body></html>' }
+
+        it 'does not raise' do
+          expect { convert }.to_not raise_error
+        end
```
abrom: Specs look good, but I think we could do better with one that also checks that the result actually was a PDF with the expected content. Something like:

```suggestion
        let(:url_or_html) { '<html><body style="background-color: blue">puppeteer-core works!</body></html>' }

        it { expect { convert }.not_to raise_error }
        it { expect(pdf_text_content).to eq 'puppeteer-core works!' }
```

spec/grover/processor_spec.rb
```diff
@@ -95,6 +101,23 @@
         end
       end
 
+      context 'when puppeteer-core package only is installed', :remote_browser do
+        before do
+          FileUtils.move 'node_modules/puppeteer', 'node_modules/puppeteer_temp'
+        end
+        after do
+          FileUtils.move 'node_modules/puppeteer_temp', 'node_modules/puppeteer'
+        end
```
abrom: Rubocop is either going to want a newline between these, or use the inline block form:
```suggestion
        before { FileUtils.move 'node_modules/puppeteer', 'node_modules/puppeteer_temp' }
        after { FileUtils.move 'node_modules/puppeteer_temp', 'node_modules/puppeteer' }
```

abrom (approved):
Nice one @sbounmy looks great! I've flagged a few minor things but I've pushed a commit to address them 😄

abrom:
Released in [v1.1.7](https://rubygems.org/gems/grover/versions/1.1.7)

PR #230, opened 2024-03-09
Better handle WS connection cleanup

abrom:


abrom:
Based on work from @dexterouschen in #183

Issue #80, opened 2020-09-17
Grover::JavaScript::Error (Navigation failed because browser has disconnected!)


ckhall:
on Heroku, receiving this error when calling `.to_pdf`.

Followed all the instructions on setting up puppeteer on heroku.  Any ideas here?

abrom:
Going to need a lot more information.

1. have you followed all of the Heroku setup instructions? https://github.com/Studiosity/grover#running-on-heroku
2. what is your code setup before the call to `.to_pdf`?
3. what versions of things have you got installed? Grover? Puppeteer? Chromium?
4. what have you tried to figure this out?
5. have you tried searching the web for the error? I found many results. It may be that someone else has had this issue!

ckhall:
@abrom 

1. yes, to the letter.
2.  ```
        html = render_to_string('report.html.erb', layout: false)
        pdf = Grover.new(
          html,
          height: '11in',
          width: '8.5in',
          landscape: true,
          margin: {
            top: '0px',
            bottom: '0px',
            left: '0px',
            right: '0px'
          }
        ).to_pdf
```
3. grover v0.13.1, puppeteer v5.2.1, chromium ? (as i understand, puppeteer downloads latest?)
4. google "Navigation failed because browser has disconnected!" - nothing i have come across pertains to ruby or heroku.
5. see #4.

ckhall:
chromium version is `Chromium 86.0.4240.0 (r800071)` according to puppeteer changelog

ckhall:
ok, figured it out.  had to remove jemalloc buildpack.  `https://github.com/gaffneyc/heroku-buildpack-jemalloc.git` specifically and not start processes using `jemalloc.sh`. may want to add a note to the readme for Heroku section.

abrom:
I'll happily accept a PR ;)

kladaFOX:
> ok, figured it out. had to remove jemalloc buildpack. `https://github.com/gaffneyc/heroku-buildpack-jemalloc.git` specifically and not start processes using `jemalloc.sh`. may want to add a note to the readme for Heroku section.

Thanks MAN

mikkelwf:
It's an issue that this doesn't work with jemalloc enabled

Fire-Dragon-DoL:
Oh wow, I think it might be worth mentioning in the README. I just lost 2 days on this and I would have not have considered grover at all if I knew I couldn't use jemalloc (yes, it saves so much memory).

Thank you for opening this issue, I would have kept searching forever.

PR #240, opened 2024-06-13
Enable passing environment variables to the node process

jukra:
We noticed that Chrome does not behave well with jemalloc enabled, and we prefer to use that in general. 

Instead of disabling jemalloc in the whole environment, we opted to just disable it for the node process used. Based on quick search, others have had the same issue as well (https://github.com/Studiosity/grover/issues/80)

We disabled it by passing a new environment variable for the node process like this

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,16 @@ respond_to do |format|
 end
 ```
 
+#### Setting custom environment variable for node
+The `node_env_vars` configuration option enables you to set custom environment variables for the spawned node process. For example you might need to disable jemalloc in some environments (https://github.com/Studiosity/grover/issues/80).
+
+```ruby
+# config/initializers/grover.rb
+Grover.configure do |config|
+  config.node_env_vars = { "LD_PRELOAD" => "" }
+end
+```
+
 ## Middleware
 Grover comes with a middleware that allows users to get a PDF, PNG or JPEG view of
 any page on your site by appending .pdf, .png or .jpeg/.jpg to the URL.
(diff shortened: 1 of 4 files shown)


abrom (commented):
At face value this certainly seems reasonable, unfortunately it is also possible to use this change to allow a pretty broad set of attacks on the host via the middleware. Options can be parsed from the body of the page being converted, and if the system supports file uploads and has the middleware enabled to allow conversion of responses it would be possible to have the end user alter these settings per request. This might include things like including extra files (`--require ..`) or changing the default CA, or enabling process inspection, changing the node path. This is a surefire way to allow potential back-dooring or file exfiltration. 

https://nodejs.org/api/cli.html#environment-variables_1

This would either require an absolutely bullet proof set of guards in `Grover::OptionsBuilder#meta_options` or configuring this outside the bounds of the `options` passed to the Grover initialiser.

jukra:
Ah yes, you are right, I did not consider the middleware usage as we are not using it. Could this be under the config like this? So not using the options but adding one more flag here [lib/grover/configuration.rb](https://github.com/Studiosity/grover/blob/main/lib/grover/configuration.rb)

```
# config/initializers/grover.rb
Grover.configure do |config|
  config.node_env_vars = { 'LD_PRELOAD': '' }
end
```

Then passing it along for the Processor?

abrom:
Yes, exactly.. much safer

jukra:
Changed now so that `node_env_vars` is part of the config.

abrom:
Released in [v1.1.8](https://rubygems.org/gems/grover/versions/1.1.8)

PR #244, opened 2024-06-27
Use Page.setDefaultTimeout with `timeout` option

abrom:
The `Page.setDefaultTimeout` will apply the more general `timeout` option more generally 😄

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,13 +69,9 @@ jobs:
 
       - name: Run tests
         run: bundle exec rspec
-        env:
-          PUPPETEER_VERSION: ${{ matrix.puppeteer-version }}
 
       - name: Run tests with remote browser
         run: bundle exec rspec --tag remote_browser
-        env:
-          PUPPETEER_VERSION: ${{ matrix.puppeteer-version }}
 
       - name: Test & publish code coverage
         uses: paambaati/codeclimate-action@v3.2.0
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -80,6 +80,12 @@ const _processPage = (async (convertAction, urlOrHtml, options) => {
 
     page = await browser.newPage();
 
+    // Apply global page timeout
+    const timeout = options.timeout; delete options.timeout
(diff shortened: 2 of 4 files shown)


jadehamel (commented):
Some test suggestions: to ensure that edge cases are covered and the global timeout interacts correctly with other options and operations. It includes scenarios where the global timeout interacts with specific options like `waitForSelector ` or when performing multiple asynchronous operations. 

```
describe 'Timeout handling with waitForSelector' do
  let(:url_or_html) do
    <<-HTML
      <html>
        <body></body>
        <script>
          setTimeout(function() {
            document.body.innerHTML = '<h1>Hey there</h1>';
          }, 200);
        </script>
      </html>
    HTML
  end

  let(:options) { basic_header_footer_options.merge('timeout' => timeout, 'waitForSelector' => 'h1') }

  context 'when the timeout is less than the page content load' do
    let(:timeout) { 100 }

    it_behaves_like 'raises navigation timeout error', timeout: 100
  end

  context 'when the timeout is greater than the page content load' do
    let(:timeout) { 2000 }

    it { is_expected.to start_with "%PDF-1.4\n" }
    it { expect(pdf_text_content).to eq "#{date} Hey there #{protocol}://www.example.net/foo/bar 1/1" }
  end
end

describe 'Handling multiple async operations within global timeout' do
  let(:timeout) { 3000 } # 3 seconds
  let(:options) { basic_header_footer_options.merge('timeout' => timeout) }

  let(:url_or_html) do
    <<-HTML
      <html>
        <body>
          <img src="large-image.jpg" />
          <script>
            setTimeout(function() {
              document.body.innerHTML += '<h1>Additional Content</h1>';
            }, 2000);  // 2 seconds delay
          </script>
        </body>
      </html>
    HTML
  end

  it 'completes all operations within the global timeout period' do
    expect(convert).to start_with "%PDF-1.4\n"
    expect(pdf_text_content).to include "Additional Content"
  end

  context 'when global timeout is exceeded by operations' do
    let(:timeout) { 1000 } # 1 second

    it_behaves_like 'raises global timeout error', timeout: 1000
  end
end

```

PR #246, opened 2024-07-11
fix: introduce ignoreDefaultArgs into launchParams

tomaswitek:
I encountered an issue when using German hyphens on Linux.

Snímek obrazovky 2024-07-11 v 13 34 02

The following images were generated with identical code on different environments.
The left side is Linux, and the right side is MacOS.
Here is a bug report on the Chromium website: https://issues.chromium.org/issues/40933841
There is a workaround for Puppeteer:
```
puppeteer.launch({
  ignoreDefaultArgs: ['--disable-component-update'],
})
```

Unfortunately,`ignoreDefaultArgs` are not passed via Grover. 
This PR fixes it.

### DIFF
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -74,6 +74,11 @@ const _processPage = (async (convertAction, uriOrHtml, options) => {
         launchParams.executablePath = executablePath;
       }
 
+      // ignoreDefaultArgs
+      if (options.ignoreDefaultArgs) {
+        launchParams.ignoreDefaultArgs = options.ignoreDefaultArgs;
+      }
+
       // Launch the browser and create a page
       browser = await puppeteer.launch(launchParams);
     }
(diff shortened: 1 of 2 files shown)


abrom (changes requested):
Thank you for the PR @tomaswitek ! 

The direction is looking good but I've flagged a few things I think could do with some tweaks.

I believe the `OptionsFixer#fix_array_options!` will also need to be updated to include this new option. Although AFAICT the expected types are slightly different in that the launch args should be an array of strings, whereas the `ignoreDefaultArgs` can be either an array of strings, or a boolean (`false`) to ignore ALL defaults. It *may* need a special handler to deal with that properly.. although I don't think so.

It would also be great to see an update to the README to explain how this can be used. I'd suggest this would apply everywhere the `launch_args` option is described. I'd also suggest referencing back to the puppeteer documentation: https://pptr.dev/api/puppeteer.launchoptions

Thoughts?

Issue #252, opened 2024-08-07
Undefined method `#pack` for nilClass on latest puppeteer release (23.0.0)


taylorthurlow:
puppeteer 23.0.0 looks to have broken Grover.

I was able to fix the issue by changing line 24 in `lib/grover/processor.rb` from:

```ruby
result['data'].pack('C*')
```

to:

```ruby
result.values.pack('C*')
```

The `result` value was expected to be a hash with a `'data'` key, and now it appears to be a hash between a String and an integer (one example k/v pair is `"135890"=>52`). It's not immediately obvious where the breaking change was that causes this, [perhaps it is this PR?](https://github.com/puppeteer/puppeteer/pull/12823)

rud:
!!I'm just a rando using the library!!

@taylorthurlow I know it's just a single line change, but would you mind making it into a patch, along with a comment for the README that puppeteer **has** to be 23.x to work with the next release of `grover`?

Kay-pet:
Hi,

I am on 

puppeteer 23.1.0

grover 1.1.9

node v18.16.1

And i am getting this same error even though I locally made the change in `processor.rb` on line 24 to `result["data"]&.pack("C*") || result.values.pack("C*")` line you have in the [open PR](https://github.com/Studiosity/grover/pull/256).

```
[5] pry(main)> grover = Grover.new('<html><p>Foo</p></html>')
#<Grover:0x102580 @uri="<html><p>Foo</p></html>">
[6] pry(main)> grover.to_pdf
NoMethodError: undefined method `pack' for nil:NilClass
from /Users/kayla.peters/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/grover-1.1.9/lib/grover/processor.rb:24:in `convert'
```

Just curious if you have any insight on why this is still failing after I made the change to `result["data"]&.pack("C*") || result.values.pack("C*")` since this is just in the rails console.

Let me know if you need more information. Thank you

taylorthurlow:
@Kay-pet Hard to say exactly, would be easier to tell if we knew what `result` is in your case. I wouldn't expect the first half of the double-pipe-or expression to raise the exception because of the safe navigation operator so my best guess at this point would be that `result.values` is returning `nil`, but why that would be the case I'm not sure.

Best thing would probably be to stick a debugger in there and inspect the variables there to see what value isn't as it is expected to be.

Kay-pet:
[...]

PR #256, opened 2024-08-27
Fix: handle puppeteer result format for v23

lorenzk:
Fixes https://github.com/Studiosity/grover/issues/252

I just took @taylorthurlow's fix and made it work with older versions of puppeteer. Tested locally with puppeteer 23 and 22.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,8 @@ jobs:
           '19.11.1',
           '20.9.0',
           '21.9.0',
-          '22.12.0'
+          '22.12.0',
+          '23.2.1'
         ]
         include:
           - ruby-version: '2.7'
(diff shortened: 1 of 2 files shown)


andrew-smith-nz (approved):
Made this same change myself manually before finding this PR.  Can confirm it works.

rud (approved):
Awesome! Would it be possible to add a testcase to verify the behaviour?

lorenzk:
I took a look at the specs but did not see a way to manipulate puppeteers output. I believe the real test is running `processor_spec.rb` against puppeteer 23.x, so I added that to the test matrix.

If you cherry-pick just the test matrix change to `main`, you should see it red and then green on this branch.

abrom:
Thanks for the PR @lorenzk 👍 

Note i've patched this "upstream" in the `processor.tsx` script instead to keep the Ruby interface unified (see #258)

The fix has been released in [v1.1.10](https://rubygems.org/gems/grover/versions/1.1.10)

PR #258, opened 2024-09-01
Add support for Puppeteer v23.x

abrom:
Puppeteer 23.x has switch from returning a `Buffer` to returning `Uint8Array` which has broken the result parsing in Grover. This wraps the convert action with `Buffer.from` (supports both `Buffer` and `Uint8Array`) to return a `Buffer` in a consistent way.


abrom:
Closes #252 and supersedes #256

PR #262, opened 2024-10-01
Attempt to use pnp strategy to find puppeteer

bradherman:
If using Yarn 3+, the default strategy is [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp).  The Grover gem errors on initialization because it can't find the puppeteer(-core) library, given there is no `node_modules` folder... This modification attempts to find `puppeteer` via `.pnp.cjs`

### DIFF
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -1,4 +1,42 @@
-// Setup imports
+const fs = require('fs')
+const path = require('path')
+
+function findPnpFile(startDir) {
+  let currentDir = startDir;
+  while (currentDir !== path.parse(currentDir).root) {
+      const pnpPath = path.join(currentDir, '.pnp.cjs');
+      if (fs.existsSync(pnpPath)) {
+          return pnpPath;
+      }
+      currentDir = path.dirname(currentDir);
+  }
+  return null;
+}
+
+function checkPuppeteerInstalledWithPnp() {
+  const pnpPath = findPnpFile(process.cwd());
+
+  if (!pnpPath) {
+    throw('Error: .pnp.cjs file not found.');
+  }
+
+  try {
+    require(pnpPath).setup();
+    try {
+      require('puppeteer');
+    } catch (error) {
+      try {
+        require('puppeteer-core');
+      } catch (error) {
+        throw (error);
+      }
+    }
+    return true;
+  } catch (error) {
+    throw (error);
+  }
+}
+
 try {
   const Module = require('module');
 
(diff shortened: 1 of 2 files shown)


lib/grover/js/processor.cjs
```diff
@@ -1,4 +1,42 @@
...
+}
+
+function checkPuppeteerInstalledWithPnp() {
+  const pnpPath = findPnpFile(process.cwd());
+
+  if (!pnpPath) {
+    throw('Error: .pnp.cjs file not found.');
+  }
+
+  try {
```
abrom, line 23: wrapping this in a try/catch (that is then just re-thrown) seems unnecessary?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "grover",
-  "version": "1.1.10",
+  "version": "1.1.11",
```
abrom, line 3: Generally best practice to leave version numbers as is and let the package maintainer worry about that 😄 

Not too much an issue with this PR, but for larger projects it would tend to lead to commit conflicts or potential issues with general release policy (ie should it be a patch release or a minor release etc).

lib/grover/js/processor.cjs
```diff
@@ -1,4 +1,42 @@
-// Setup imports
+const fs = require('fs')
+const path = require('path')
+
+function findPnpFile(startDir) {
+  let currentDir = startDir;
+  while (currentDir !== path.parse(currentDir).root) {
+      const pnpPath = path.join(currentDir, '.pnp.cjs');
+      if (fs.existsSync(pnpPath)) {
```
abrom, line 8: although the likelihood of this being a TOCTOU issue is low, I think it'd still be worth refactoring to avoid using any `exists` methods. ie 
1. calculate the path (the line above)
2. use it by calling `require(pnpPath).setup();` (wrapped in a try/catch)
   - if it succeeds, well.. you know the path was good and we can just try require puppeteer/core
   - if it didn't succeed then we try the next path

It would require flipping the logic a little on its head, but something like this would appear to do it. Note, I don't have yarn 4 on my system so I haven't tested this 😬. Speaking of testing, it would be good if there was a way for this to be tested as a part of the CI so as to avoid regressions. I'm keen to hear what thoughts you might have on this 😄 

```
function checkPuppeteerInstalledWithPnp() {
  let currentDir = process.cwd();
  while (currentDir !== path.parse(currentDir).root) {
    const pnpPath = path.join(currentDir, '.pnp.cjs');
    try {
      require(pnpPath).setup();
      try {
        require('puppeteer');
      } catch (error) {
        require('puppeteer-core');
      }
      return;
    } catch (loadError) {
      // That didn't work.. lets try the next path
    }
  }
  throw('Error: .pnp.cjs file not found.');
}
```

abrom (changes requested):
thanks for the PR @bradherman 👍 

I've made a few comments about some areas I think we could improve, and I'm keen to hear what you think about how this might be tested as a part of the CI system.

lib/grover/js/processor.cjs
bradherman: will look into this in a few days... on vacation for the last few weeks... also worth noting i'm super new to any JS stuff, so I'm learning here!  thanks for the feedback and I plan to address when I'm back home :pray:

abrom: no worries, enjoy the holiday 😄

le0pard:
Possible simple solution for this - just ask where `puppeteer` from `yarn`, like:

```bash
$ yarn bin puppeteer
/Users/leo/programs/project/.yarn/unplugged/puppeteer-npm-23.7.0-1916e36ee4/node_modules/puppeteer/lib/cjs/puppeteer/node/cli.js
```

abrom:
FYI @bradherman this should now be resolved with [v1.2.0](https://rubygems.org/gems/grover/versions/1.2.0)

PR #265, opened 2024-11-03
Unwrap hash in Grover.new examples

sabslikesobs:
Some `Grover.new` examples use (an old?) format for invocation where an explicit hash is the second argument. These fail because Grover expects keyword arguments.


abrom (approved):
thanks @sabslikesobs 👍

PR #275, opened 2024-12-30
Add Ruby 3.4 support

### LINKED ISSUE #273: Ruby 3.4 error
Hello, have an error
image

The same code works fine with Ruby 3.3.6. No issues. PDF is simple, nothing special.

If I debug I see the following in "result" variable

image

murajun1978:
Support Ruby 3.4

Fixed #273


Laykou:
@abrom Could this be merged and released in rubygems please? Thank you!

grover.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |spec|
   SUMMARY
   spec.homepage    = 'https://github.com/Studiosity/grover'
   spec.license     = 'MIT'
-  spec.required_ruby_version = ['>= 3.0.0', '< 3.4.0']
+  spec.required_ruby_version = ['>= 3.0.0', '< 3.5.0']
```
jibiel, line 19: @abrom Any particular reason to lock Ruby version so tightly in this gem?

It'll be the same problem a few months later when 3.5 is released.
The gem prevents apps using it from upgrading Ruby.

pboling: I agree, it is not a reasonable restriction, except in the school of enterprisey business decisions.  Please relegate it to the foibles of the Ruby ecosystem's past!
This is blocking upgrades for me, so I may switch to sourcing from a git fork.

erichstark:
Any updates on this?

pboling (changes requested):
Do not restrict to less than ruby 3.5 unless you want this gem to once again be a major upgrade blocker in 11 months.

grover.gemspec
abrom: > @abrom Any particular reason to lock Ruby version so tightly in this gem?
> 
> It'll be the same problem a few months later when 3.5 is released. The gem prevents apps using it from upgrading Ruby.

Because it hasn't been tested

abrom: > I agree, it is not a reasonable restriction, except in the school of enterprisey business decisions. Please relegate it to the foibles of the Ruby ecosystem's past! This is blocking upgrades for me, so I may switch to sourcing from a git fork.

Not reasonable to you, but expected by anyone who needs gems to work with the versions they're indicated as being supported by. 

I agree that it is a blocker, no doubt about it. The Ruby versions releasing over what is traditionally a holiday for many, including myself, shouldn't dictate the entire community drop everything to support a pretty extreme minority who need instant version bumps 😉 . If you absolutely need to be using the latest Ruby as soon as it's released is going to result in the need, as you've suggested, for using forks.. that's a given.

abrom:
> Do not restrict to less than ruby 3.5 unless you want this gem to once again be a major upgrade blocker in 11 months.

You want the gem to support a yet to be released version of the language ? In all likelihood it would work.. yes.. probably. But sorry, I'm not going to open the release versions up for a probably. Each release runs through a CI suite to ensure the gem is going to behave *exactly* how it is supposed to. How would that work for a yet to be released language? I can assure you that the vitriol I would receive because the gem SAID it supports Ruby X.Y and then for some reason it didn't.. *sigh*

grover.gemspec
benjamineskola: @abrom I get where you're coming from, but most gems don't seem to include a maximum supported Ruby version in their gemfiles from what I can tell, and for the most part this doesn't seem to lead to any problems. Is there a particular concern that grover is likely to break on new versions?

(Grover isn't the only one — I think I noticed nokogiri did too. But they released a new version the same day. I'm absolutely not going to suggest that you should be expected to release new versions on Christmas Day; but I do wonder if removing the upper bound would lead to any actual problems.)

abrom:
Thanks @murajun1978 👍 Released in [v1.2.2](https://rubygems.org/gems/grover/versions/1.2.2)

grover.gemspec
abrom: image

😉

pboling:
> Each release runs through a CI suite to ensure the gem is going to behave exactly how it is supposed to. How would that work for a yet to be released language?

I can only tell you how I do it for the dozens of gems I maintain, and that is: I run the CI against `ruby-head`, which is the next Ruby release.

PR #294, opened 2025-07-28
Use the browser object to set cookies. 

brendon:
It appears that using the page to set cookies is now deprecated.

I'm not sure if this will change the dependency graph of the gem as I'm not sure when this was actually changed, or if it's always been allowed:

https://pptr.dev/guides/cookies

### DIFF
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -114,7 +114,7 @@ const _processPage = (async (convertAction, uriOrHtml, options) => {
     // Setting cookies
     const cookies = options.cookies; delete options.cookies
     if (Array.isArray(cookies)) {
-      await page.setCookie(...cookies);
+      await browser.setCookie(...cookies);
     }
 
     // Set caching flag (if provided)


abrom (changes requested):
Good catch, thanks @brendon 👍 

However per my comment below I think we could handle this a little better to support the deprecated form, at least in the short term.

brendon:
On a mildly related note, I spent a few hours trying to figure out why my coping the Rails session cookie into puppeteer wasn't working. Turns out you'd already solved the problem in your middleware and the solution is to escape the value first: `Rack::Utils.escape(value)`.

Do you think this would be a good thing to add to the readme in the section that deals with cookie transfer?

abrom:
Sorry @brendon I missed your comments on this PR 😢 

> Do you think this would be a good thing to add to the readme in the section that deals with cookie transfer?

Sure, sounds like a good idea

abrom (approved):

abrom:
FYI this was released in [v1.2.4](https://rubygems.org/gems/grover/versions/1.2.4)

PR #305, opened 2026-03-26
Fix session isolation for remote browser WebSocket connections

wooly:
When using `browser_ws_endpoint` to connect to a remote Chrome instance, pages are created via `browser.newPage()` in the default browser context. Cookies (and other browsing state) set during one request persist in the default context and leak into subsequent requests from other users.

This PR updates the behaviour to use `browser.createBrowserContext()` instead. This behaves like an incognito session — state is fully isolated and automatically cleared when `context.close()` is called after the request completes.

Added a test that:
1. Launches a persistent Chrome instance (simulating a remotely running Chrome) and retrieves its WS endpoint
2. Makes a first request that sets a cookie via `evaluateOnNewDocument`
3. Makes a second request with no cookies and asserts the cookie from the first request is not visible

Without the fix, the second request sees `session-secret=abc123` (bleed). With the fix it sees `no cookies`.

### DIFF
--- a/lib/grover/js/processor.cjs
+++ b/lib/grover/js/processor.cjs
@@ -34,7 +34,7 @@ function GroverError(name, errors) {
 GroverError.prototype = Error.prototype;
 
 const _processPage = (async (convertAction, uriOrHtml, options) => {
-  let browser, page, tmpDir, wsConnection = false;
+  let browser, context, page, tmpDir, wsConnection = false;
   const requestErrors = [], pageErrors = [];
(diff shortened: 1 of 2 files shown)


abrom (approved):
A few Rubocop lints that needed addressing, but otherwise looks great. Nice one @wooly 👍

abrom:
Released in [v1.2.10](https://rubygems.org/gems/grover/versions/1.2.10)

